PR #2, opened 2022-08-17
Sanitize back-ticks in strings when there is a possibility of Cypher injection

gem-neo4j:
Back-ticks will be escaped to allow users to still have back-ticks in their labels, properties etc as in Cypher, without the risk of Cypher Injection.

This also fixes this issue: https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3139

### DIFF
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -61,6 +61,8 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
@@ -670,8 +672,19 @@ public class Util {
         return SourceVersion.isIdentifier(var) && !var.contains("$") ? var : '`' + var + '`';
     }
 
-    public static String sanitizeAndQuote(String var) {
-        return quote(var.replaceAll("`", ""));
+    // Escape all backticks in a string to stop Cypher Injection Attacks
+    public static String sanitizeBackTicks(String text) {
+        Pattern finOddCountBackticks = Pattern.compile("(?<!`)`(?:`{2})*(?!`)");
+        Matcher matcher = finOddCountBackticks.matcher(text);
+
+        int currentMatch = 0;
+        String newText = text;
+        while (matcher.find()) {
+            newText = new StringBuilder(newText).insert(matcher.start() + currentMatch, "`").toString();
+            currentMatch++;
+        }
+
+        return newText;
     }
 
     public static String param(String var) {
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -672,23 +672,53 @@ public class Util {
         return SourceVersion.isIdentifier(var) && !var.contains("$") ? var : '`' + var + '`';
     }
 
-    // Escape all backticks in a string to stop Cypher Injection Attacks
-    public static String sanitizeBackTicks(String text) {
-        if (text == null) return null;
-        // Replace unicode backticks with the backtick char first
-        text = text.replaceAll("\u0060", "`");
-        Pattern findOddCountBackticks = Pattern.compile("(?<!`)`(?:`{2})*(?!`)");
-        Matcher matcher = findOddCountBackticks.matcher(text);
+    private static final String BACKTICK_OR_UC = "[`\\\\\u0060]";
 
-        int currentMatch = 0;
-        String newText = text;
-        while (matcher.find()) {
-            newText = new StringBuilder(newText).insert(matcher.start() + currentMatch, "`").toString();
-            currentMatch++;
+    private static final Pattern LABEL_AND_TYPE_QUOTATION = Pattern.compile(
+            String.format("(?<!%1$s)%1$s(?:%1$s{2})*(?!%1$s)", BACKTICK_OR_UC));
+
+    /**
+     * This is a literal copy of {@code javax.lang.model.SourceVersion#isIdentifier(CharSequence)} included here to
+     * be not dependent on the compiler module.
+     *
+     * @param name A possible Java identifier
+     * @return True, if {@code name} represents an identifier.
+     */
+    public static boolean isIdentifier(CharSequence name) {
+        String id = name.toString();
+
+        if (id.length() == 0) {
+            return false;
+        }
+        int cp = id.codePointAt(0);
+        if (!Character.isJavaIdentifierStart(cp)) {
+            return false;
+        }
+        for (int i = Character.charCount(cp); i < id.length(); i += Character.charCount(cp)) {
+            cp = id.codePointAt(i);
+            if (!Character.isJavaIdentifierPart(cp)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


gem-neo4j:
[APOC Backtick Vulnerability](https://trello.com/c/1zjbkDMC/248-apoc-backtick-vulnerability)

common/src/main/java/apoc/util/Util.java
```diff
@@ -670,8 +672,19 @@ public static String quote(String var) {
         return SourceVersion.isIdentifier(var) && !var.contains("$") ? var : '`' + var + '`';
     }
 
-    public static String sanitizeAndQuote(String var) {
-        return quote(var.replaceAll("`", ""));
+    // Escape all backticks in a string to stop Cypher Injection Attacks
+    public static String sanitizeBackTicks(String text) {
+        Pattern finOddCountBackticks = Pattern.compile("(?<!`)`(?:`{2})*(?!`)");
+        Matcher matcher = finOddCountBackticks.matcher(text);
```
AzuObs, line 678: image

AzuObs, line 678: https://media2.giphy.com/media/13Qd06UWpuicJW/giphy.gif

AzuObs, line 678: https://media2.giphy.com/media/13Qd06UWpuicJW/giphy.gif

AzuObs: Looks good but could we add some tests for this to make sure we covered all cases and to prevent regressions in the future?

core/src/main/java/apoc/refactor/GraphRefactoring.java
```diff
@@ -552,11 +552,14 @@ private Future<Void> categorizeNodes(List<Node> batch, String sourceKey, String
             node = Util.rebind(innerTx, node);
             Object value = node.getProperty(sourceKey, null);
             if (value != null) {
+                String nodeLabel = Util.sanitizeBackTicks(label);
+                String key = Util.sanitizeBackTicks(targetKey);
+                String relType = Util.sanitizeBackTicks(relationshipType);
```
AzuObs, line 557: From apoc.refactor.categorize it looks like `label` could be null. I didn't check the others. You may need to handle this here, or in your sanitize function to avoid NPEs

AzuObs: There seems to be other functions for which this might an issue.

core/src/test/java/apoc/refactor/rename/RenameTest.java
```diff
@@ -55,6 +55,55 @@ public void testRenameLabel() throws Exception {
...
+		db.executeTransactionally("UNWIND range(0,9) AS id CREATE (f:Foo {id: id})");
+		testCall(db, "CALL apoc.refactor.rename.label(" +
+						"  'Foo', " +
+						"  'Whatever` WITH n MATCH (m) DETACH DELETE m //'" +
+						")",
+				map(),
+				(r) -> {});
+
+		assertEquals(10L, resultNodesMatches("`Whatever`` WITH n MATCH (m) DETACH DELETE m //`", null));
+		assertEquals(0L, resultNodesMatches("Foo", null));
```
AzuObs, line 69: It seems that there we are wrapping the entire label in more "`" than needed.

```
CALL apoc.refactor.rename.label("Foo", "Whatever")  // renames label "Foo" to "Whatever"
CALL apoc.refactor.rename.label("Foo", "What`ever") // rename label "Foo" to "What`ever", not "`What`ever`"
```

Thus in your test I would expect 
```
resultNodesMatches("`Whatever`` WITH n MATCH (m) DETACH DELETE m //`", null)); //  == 0
```

Apologies if I'm missing something, please feel free to reach out on Slack

AzuObs (commented):
Great work so far. Left a couple of nitpicks. It's incredible that we have this much code vulnerable!

It's very hard for me to review whether we caught all the vulnerabilities, and I don't really have the luxury of doing a complete search myself. Would you mind sharing your methodology for finding vulnerable code, so that we may asses its completeness?

core/src/test/java/apoc/refactor/rename/RenameTest.java
gem-neo4j: Oh, no it is just the test function resultNodesMatches, didn't add wrapper backticks so this would have tried to find nodes called `node.Whatever`` WITH  n MATCH (m) DETACH DELETE m //` which isn't valid cypher (the spaces break it), would you prefer me to change the tester function to add backticks everywhere for consistency?

common/src/main/java/apoc/util/Util.java
```diff
@@ -670,8 +672,22 @@ public static String quote(String var) {
         return SourceVersion.isIdentifier(var) && !var.contains("$") ? var : '`' + var + '`';
     }
 
-    public static String sanitizeAndQuote(String var) {
-        return quote(var.replaceAll("`", ""));
+    // Escape all backticks in a string to stop Cypher Injection Attacks
+    public static String sanitizeBackTicks(String text) {
+        if (text == null) return null;
+        // Replace unicode backticks with the backtick char first
+        text = text.replaceAll("\u0060", "`");
```
michael-simons, line 679: This basically resolved to `text.replaceAll("`", "`")`

common/src/main/java/apoc/util/Util.java
```diff
@@ -670,8 +672,22 @@ public static String quote(String var) {
...
+        Pattern findOddCountBackticks = Pattern.compile("(?<!`)`(?:`{2})*(?!`)");
```
michael-simons, line 680: Should go to a static final var, slow to compile.

gem-neo4j:
@AzuObs In response to your question about how I searched for vulnerabilities. Firstly, with the naive way of just searching where we add backticks around user code in the code base. Then, I basically, just used the docs list of procedures and went through all the ones that have input that could be suspicious (e.g string, object) and checked out their code, so most of the ones that look suspicious are using the nice in built functions, which handle this for us e.g node.addLabel(). It was a bit manual, but felt like the most thorough way for me to do it.

core/src/test/java/apoc/refactor/rename/RenameTest.java
AzuObs: Oh sorry that was my confusion then! No need to change it

AzuObs (approved):
Great job!

core/src/test/java/apoc/util/UtilTest.java
```diff
@@ -89,6 +89,30 @@ public void testValidateQuery() {
...
+        assertEquals("Foo `\\u0060", Util.sanitizeBackTicks("Foo \\u0060"));
+        assertEquals("ABC", Util.sanitizeBackTicks("ABC"));
+        assertEquals("A C", Util.sanitizeBackTicks("A C"));
+        assertEquals("A`` C", Util.sanitizeBackTicks("A` C"));
+        assertEquals("ALabel", Util.sanitizeBackTicks("ALabel"));
+        assertEquals("A Label", Util.sanitizeBackTicks("A Label"));
+        assertEquals("A ``Label", Util.sanitizeBackTicks("A `Label"));
+        assertEquals("``A ``Label", Util.sanitizeBackTicks("`A `Label"));
+        assertEquals("``A ``Label", Util.sanitizeBackTicks("`A `Label"));
+        assertEquals("Spring Data Neo4j⚡️RX", Util.sanitizeBackTicks("Spring Data Neo4j⚡️RX"));
```
michael-simons, line 112: <3

michael-simons:
Great job, nice testing added too. Thank you.

michael-simons (approved):

PR #6, opened 2022-08-23
Remove APOC Extended from Docs

gem-neo4j:
A lot of deletion 🙃 if you want to see the docs instead do the following from the repo:

cd docs/antora
npm update
npm start

Then open index.html from docs/antora/build/site/apoc/5/index.html

(Note: my Intellij didn't like me doing one then the other, so I had to restart and invalidate caches to get it to work in the other repo)


gem-neo4j:
[Split apoc docs](https://trello.com/c/uGTIvlKp/253-split-apoc-docs)

AzuObs:
5/introduction/index.html 

Out of scope for this PR because I think we ought to discuss this as a team and it should go into a separate ticket, but the introduction page could perhaps introduction a section about the significant breaking change happening in 5.0

AzuObs:
5/installation/index.html

Screenshot 2022-08-25 at 11 40 22

All Github links in this project should probably be updated as part of this ticket, but let me know if you agree with that. Links which are not Github links probably can't be updated right now.

AzuObs:
5/usage/index.html

Maybe this link could also be updated? We already have some sort of Github site hosted (example page: https://neo4j.github.io/apoc/versions.json)

Screenshot 2022-08-25 at 11 48 56

AzuObs:
Just a remark.

Lots of videos in here which have content containing APOC Full. Would be interesting to see what happens with these in the future. IMO we ought to consider removing them, but decision should involve more people and is out of scope of this PR.

AzuObs:
5/overview/index.html

I started to go through the list of APOC Full function removed from this page. Looks good.

Unfortunately there are still references to these functions in other parts of the documentation. Would you agree that splitting the docs should mean removing all references to these?

I didn't go much further from the letter "d", but there are more than these examples:
- apoc.cluster.graph
- apoc.custom.*
- apoc.cypher.*
- apoc.data.email
- apoc.date.expire
- apoc.date.expireIn

gem-neo4j:
@AzuObs thanks for the comments :) I agree about the links, will update there!

I'll double check through all the procedures, I didn't realise I had missed so many! I think some of these are just in the deprecations-and-additions page which I left as is, but you're right, that probably shouldn't have any in there either :)

AzuObs:
In `docs/img` there seem to be some assets which are only used by Extended and not by core like `docs/img/apoc.bolt.*` for example. These should probably be removed (while checking there are no links to them in the docs).

AzuObs:
In `img/docs` there seem to be some assets which contain content that is not in APOC Core. `docs/img/apoc-help.jpg` for example contains some text which mentions `apoc.load.driver` which is part of Extended.

I'm not sure if the images will be migrated and will be supported in the new docs which we will collaborate with the docs team on. But if they intend to keep these images then I would probably suggest that we modify/recapture them.

AzuObs:
In the Deprecation section it may be nice to mention something about the split. But I think this ought to be part of a larger discussion of how we will communicate the changes happening to APOC.

AzuObs:
There is a package which has historically been mislabelled in the docs: https://neo4j.com/labs/apoc/4.1/overview/apoc.dv 

In the source code they are under Extended, and in fact they are not part of this repository. I've had a chat with Nacho since he is in the office and worked on the split, and we think that maybe we should remove it from the docs rather than reintroduce it in the doc. 

If we added to the code it would be hard to deprecate later, but if we remove it from the docs and people complain then we can take more time and really think if we want to add it back to the code.

I imagine there may be other cases where the docs are wrong like this, it may be worth checking - WDYT?

AzuObs:
**REFERENCES TO neo4j-apoc-procedures** 

- 5/installation/index.html#neo4j-server

The versioning scheme we have for this project will no longer be <neo4j>.<apoc> versioning scheme which results in a semver with 4 parts (eg 4.4.0.1). That compatibility matrix probably also won't be super useful in the future.

In fact, some of this section probably needs quite a lot of modifications. Basically the versioning scheme is going to be semver, and it will be aligned with Neo4j versions but not in the way described I imagine. We will also be making the guarantee that using the latest APOC plugin releasable will just work with whatever 5+ release of Neo4j you are using thanks to our compatibility layer. Some details need to be figured out about the versioning guarantees we make, for example will we only provide backwards compatibility between minor versions?

Not sure if this should be part of this ticket or not, but it would probably be good to get this figured out before our initial release of the new docs.

- 5/import/web-apis/index.html#_using_hdfs_protocol
`cd neo4j-apoc-procedures/extra-dependencies` and `will be created into the neo4j-apoc-procedures/extra-dependencies/hadoop/build/lib folder.`

AzuObs:
Second round of review here is DONE. Well done on all the work!

gem-neo4j:
@AzuObs Ah yes, I left the versioning as is atm because I wasn't sure how it was going to look, and the information there is still true for the previous versions. I think it would make sense to have a pre and post 5.0 version here as well, WDYT? :)

PR #9, opened 2022-08-30
Adapt to changes in Coordinate class in product

MishaDemianenko:


AzuObs (approved):

ncordon (commented):
I'm unsure whether this would pass the spotless check we have. Can you run `./gradlew spotlessApply` and then re-commit if there are any changes?

Otherwise looks good

MishaDemianenko:
> ./gradlew spotlessApply

looks happy with current formatting and no changes made on local env as result of that

ncordon (changes requested):
This needs to be pushed to the repo directly, not to a fork

MishaDemianenko:
> This needs to be pushed to the repo directly, not to a fork

why?

ncordon:
> This needs to be pushed to the repo directly, not to a fork

Because the CI secrets are only available in branches in the repo and it will not be able to download in-dev neo4j artifacts in forks

PR #166, opened 2022-09-01
Remove extended APOC config from core

Lojjs:
Parts of the APOC config was only used from extended (or in some cases fully unused). This PR removes them from core, while xxx introduce corresponding config to the extended repo.


common/src/main/java/apoc/ApocConfig.java
```diff
@@ -46,23 +44,15 @@
 import static org.neo4j.configuration.GraphDatabaseSettings.transaction_logs_root_path;
 
 public class ApocConfig extends LifecycleAdapter {
-    public static final String SUN_JAVA_COMMAND = "sun.java.command";
+    protected static final String SUN_JAVA_COMMAND = "sun.java.command";
     public static final String APOC_IMPORT_FILE_ENABLED = "apoc.import.file.enabled";
     public static final String APOC_EXPORT_FILE_ENABLED = "apoc.export.file.enabled";
     public static final String APOC_IMPORT_FILE_USE_NEO4J_CONFIG = "apoc.import.file.use_neo4j_config";
     public static final String APOC_TTL_SCHEDULE = "apoc.ttl.schedule";
```
gem-neo4j, line 51: Why are some TTL configs still in core? I thought all TTL procedures were in extended

Lojjs: The reason I kept this as well as some UUID constant here was because of them being used in the file ApocSettings.java. I also tried out to move those settings to a ExtendedApocSettings file in extended, but then I would have to copy over all the code under `// copy apoc settings from neo4j.conf for legacy support` in ApocConfig to the ExtendedApocConfig as well if we want to keep the old behaviour.

I'm not sure if this legacy support is really needed in 5.x though as the comment in ApocSettings sounds a bit like that is something to make it easier for people to use the 3.x way in 4.x. If we think we can get rid of it all together, I could move over those constants as well as do further cleanup.

gem-neo4j: Ah okay, I think the main thing for me is that it feels strange to have extended settings still in core. It might be nice to see if it is possible to remove the legacy stuff so core can move these out :) maybe ask the team?

gem-neo4j (approved):
Looks good! Nice with this all cleaned up 😄

PR #167, opened 2022-09-01
Fix versions

AzuObs:
[6699d92](https://github.com/neo4j/apoc/pull/167/commits/6699d92907725b0842a6a265be2c0218875f8559) Cleanup 3.x documentation

Docs for APOC 3.x don't seem to be hosted anywhere, but even if they were hosted they would be hosted from neo4j-contrib/apoc. As such, these files can be safely removed from neo4j/apoc.

[fe8f121](https://github.com/neo4j/apoc/pull/167/commits/fe8f1210a31b17ee55f7e80619af84932ec35d4a) Update ./gradlew checksums task

The checksum task works nearly the same as it did before. The only difference is updating URL references and adding TODO to delete the task once we begin signing jars.

[be65a9a](https://github.com/neo4j/apoc/pull/167/commits/be65a9a5a0c105785edbcf2522cfb5ffb2a9242a) Update ./gradlew versions task

The versions task works mostly the same as it did before. The main difference is how we link a Neo4j version to an APOC version. The logic here used to be a bit complex because of various edge cases, but going forward APOC will guarantee that it is backwards compatible with any known major version of APOC. This guarantee greatly simplifies the job.

[f1c1455](https://github.com/neo4j/apoc/pull/167/commits/f1c14551cd84952d8449e1ef4ef4c9cb72be58ee) Commit updated checksums.json and versions.json

Commit the update checksum.json and versions.json file which were produced by the new jobs.


build.gradle
```diff
@@ -222,12 +223,4 @@ class Checksum {
     private static toHex(int b) {
         return hexChr((b & 0xF0) >> 4) + hexChr(b & 0x0F)
     }
-
-    static def main(args) {
-        def sha1sum = Checksum.getSha1sum(new File(args[0]))
-        println "file($args[0]): sha1sum($sha1sum)"
-    }
-
```
ncordon, line 230: Any reason why we deleted this?

build.gradle
```diff
@@ -183,6 +118,10 @@ task checksums <<  {
     def checksumsFile = new File("checksums.json")
     checksumsFile.write(JsonOutput.prettyPrint(JsonOutput.toJson(checksums)))
 }
+
+// We order considering that X.Y.Z-SOMETHING should be considered lower than X.Y.Z
+static def orderVersion(str) {
+    return str.split("[.-]").collect { it =~ "^\\d+\$" ? it.padLeft(4,"0") : it.replace("beta","mbeta").replace("BETA","MBETA") }.join(".") + "/"
 }
```
ncordon, line 125: It's a bit complicated to understand what this does. Also why does it replace beta with mbeta? Does this consider the release candidates lower that the non-release candidate ones?

Note there were already release candidates published in Maven central: https://mvnrepository.com/artifact/org.neo4j.procedure/apoc

build.gradle
```diff
@@ -3,155 +3,90 @@ import groovy.json.JsonSlurper
...
-            releases: releases.findAll { !(it.prerelease || it.draft)}
-    ]
+// Any major version of Neo4j should use the latest corresponding major version of APOC
+static def findApocVersion(apocReleases, neo4jRelease) {
+    def majorNeoVersion = neo4jRelease.substring(0, 1)
+    def candidates = apocReleases
+            .findAll { it.substring(0, 1) == majorNeoVersion }
+            .sort { o1, o2 -> orderVersion(o1) <=> orderVersion(o2) }[-1]
+    return candidates[-1]
 }
```
ncordon, line 42: 👍

checksums.json
```diff
@@ -1,587 +1,12 @@
...
+  "5.0.0-rc02": {
+    "sha1": "c52f7dc54762297b772036dab163d07375908631",
+    "sha256": "6b2969665f24bc27e18eb04c0c275a40b5075ffcb58aacd8e6a53c01a6e1fbff",
+    "md5": "94ec922d16e650e5c52d1b338f1e55df"
+  },
+  "5.0.0-rc01": {
+    "sha1": "54c27ef7d014a14ed1e10ca7b7c94487aea66bf2",
+    "sha256": "528d3b13ca539e598a1b84f80f3517151efa81d2ec7537ca932c531caf4b2ac0",
+    "md5": "e75d77aa76ca178ebf1a2efede673b90"
+  }
```
ncordon, line 11: Note these versions have been erased from the other json file, I suppose that's normal cause there is no release for neo4j

ncordon (commented):
Looks good. The only feedback that is a we-must-do is to check the versions and checksums file do not shrink in size to avoid what happened here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/commit/77258d3012b29168d1c23c36422ae3a46ad120fb

build.gradle
AzuObs: It didn't seem like it was being used anywhere in the code, so I deleted it. To be certain, I did a search on TeamCity build logs.

build.gradle
AzuObs: Yes it's a little tricky so let me give an explanation and a few examples. Basically this works for both Neo4j and APOC releases. Which usually look something like this:

(**neo4j**)[https://repo1.maven.org/maven2/org/neo4j/neo4j/]
- 1.2.M01
- 3.0.0-M01
- 3.1.0-BETA1      
- 3.2.0-alpha01
- 3.5.0-beta01
- 3.5.0-rc01
- 3.5.1

(**apoc**)[https://repo1.maven.org/maven2/org/neo4j/procedure/apoc/]
- 4.0.0-rc01
- 4.1.0.1

What the function does
- splits these string on "." or "-" character
- if a split is a number then left pad it up to 4 characters
- if a split is a string then it sometimes renames the string others it leaves it as-is

I'm going to remove the string renaming part, because we don't need it for our new project. But if you're curious then it's "beta" to "mbeta" so that the beta versions are listed after the "m01", "m02", etc. versions.

The examples above end up looking like this...

**neo4j** (ordered)
- 0001.0002.M01/
- 0003.0000.0000.M01/
- 0003.0001.0000.BETA1 /
- 0003.0002.0000.alpha01/
- 0003.0005.0000.beta01/
- 0003.0005.0001/
- 0003.0005.0000.rc01/

**apoc** (ordered)
- 0004.0001.0000.0001/
- 0004.0000.0000.rc01/

Ordering is done elsewhere in the code, and it just compares the string's unicode code point, character by character. It's the same way Java does it. As such, "rc" versions appear after non-"rc" versions.

build.gradle
```diff
@@ -122,6 +126,13 @@ task checksums <<  {
 static def normalizeVersion(str) {
     return str.split("[.-]").collect { it =~ "^\\d+\$" ? it.padLeft(4,"0") : it }.join(".") + "/"
 }
+
+// We have noticed in the past that the versions.json file sometimes gets wiped clean by the task when it's run in
+// TeamCity. We assume that this is due to one or several of the HTTP requests this job makes fails silently. In these
+// scenarios it is better to simply keep the old versions.json file. We detect this scenario by verifying the new
+// versions have at least as many keys as the previous versions
+static def assertVersionsGrew(previousVersions, versions) {
+    if (previousVersions.keySet().size() > versions.keySet().size()) throw new Exception("Error: versions file shrank.")
```
ncordon, line 135: Excellent! 💃

ncordon (approved):

PR #169, opened 2022-09-02
5.0 add storage engine check

gem-neo4j:
The apoc.warmup.run procedure can only be used with the currently available storage engines.

I've added a checker for checking which storage engine currently is set via config. Let me know thoughts on this! It is also possible to only have the docs updated :)

### DIFF
--- a/core/src/test/java/apoc/warmup/WarmupEnterpriseTest.java
+++ b/core/src/test/java/apoc/warmup/WarmupEnterpriseTest.java
@@ -2,8 +2,6 @@ package apoc.warmup;
 
 import apoc.util.Neo4jContainerExtension;
 import apoc.util.TestContainerUtil;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
 import org.junit.Test;
 import org.neo4j.configuration.GraphDatabaseInternalSettings;
 import org.neo4j.configuration.GraphDatabaseSettings;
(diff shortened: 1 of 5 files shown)


gem-neo4j:
[Procedure apoc.warmup.run uses internal knowledge of Record storage engine and will not work with other storage engines](https://trello.com/c/c6W74nta/12-procedure-apocwarmuprun-uses-internal-knowledge-of-record-storage-engine-and-will-not-work-with-other-storage-engines)

core/src/test/java/apoc/warmup/WarmupEnterpriseTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @AfterClass
+    public static void afterAll() {
+        session.close();
+        neo4jContainer.close();
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testWarmupIsntAllowedWithOtherStorageEngines() {
+        testCall(session, "CALL apoc.warmup.run()", (r) -> {});
```
ncordon, line 39: Two nitpicks:
* We are missing a positive test
* Does this need to be an integration test (with a docker container)? Isn't there a way to test this with a unit test?

gem-neo4j: I can add specific positive tests :)

The only way to test the negative case was to use an enterprise db, is there a way to get a test on that without docker? When I searched for setting up enterprise in other tests this was how they did it, but maybe I misunderstood? If so let me know.

gem-neo4j: That sounds good :) I can make a card for finding that out

ncordon (approved):

PR #170, opened 2022-09-05
Use DefaultDatabaseResolver to resolve default db (instead of Config)

gfx54b:
Using Config to resolve default db is correct in Community. In Enterprise:
* in 4.x it is correct in  standalone, but not in clusters (probably nobody noticed - given the troublesome past cluster vs. apoc).
* in 5.0 it is not correct. The config value represents the "initial" default database only. The topology graph knows better, as the default database is dynamically changeable.

This PR is a prerequisite for https://github.com/neo-technology/neo4j/pull/16699


gfx54b:
Closed in favor of #171

PR #172, opened 2022-09-05
Improve backtick sanitization

gem-neo4j:
Fixes:
- Too many escapes of the backslash
- No need to check for the actual ` and the unicode value
- But a need to check for the escaped unicode proper (which is literal \u0060, written down as Java String becomes \\u0060, no further escaping for the regex necessary


AzuObs (approved):
Excellent work, thanks for raising the PR!

PR #174, opened 2022-09-13
Remove docs from Core

gem-neo4j:


gem-neo4j:
[Remove Docs from APOC core](https://trello.com/c/QqHeLNRR/291-remove-docs-from-apoc-core)

ncordon (approved):
LGTM, but I think new docs files popped in the repo and that's why it has conflicts?

PR #175, opened 2022-09-13
Fix flaky tests due to SocketTimeoutException - added assertEventually

vga91:
Trello card: https://trello.com/c/FO9tIWDL/1096-essential-builds-failure-apocloadxmltesttestloadxmlfromzipbyurl-is-failing-on-a-dev-essential-build

Since the exception is thrown when we read the url https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.4/src/test/resources/testload.zip?raw=true!xml/books.xml, and similar,
we have no control to the timeout error due to slow response.

So I replaced the `testResult` and` testCall` with `assertEventually` to prevent future `SocketTimeoutException`s, 
 e.g. in https://trello.com/c/g06vJGtc/951-essential-builds-failure-apocloadxmltesttestloadxmlfromtargzbyurl-is-failing-on-a-dev-essential-build?search_id=0b3b4048-6ce6-44fd-9469-726d051c66b5


vga91:
Closed in favor of https://github.com/neo4j/apoc/pull/178

PR #176, opened 2022-09-13
Try to fix flaky testContainers

vga91:
Trello card: https://trello.com/c/9Gnr5umk/1384-essential-builds-failure-apocwarmupwarmupenterprisetesttestwarmuponenterprisestorageengine-is-failing-on-a-dev-essential-build

I was unable to replicate the error locally.

I added a retry mechanism via `withStartupAttempts(3)` hoping it can solve the problem.

I also changed `throw new RuntimeException(ex);` with `startException.addSuppressed`, because currently we catch the exception caused by `System.out.println(this.execInContainer` (which is always `java.lang.IllegalStateException: execInContainer can only be used while the Container is running`, e.g when an image is not found) but not the original `startException`, so that it should be easier to understand the original problem in case it reoccurs


vga91:
Closed in favor of https://github.com/neo4j/apoc/pull/177

PR #178, opened 2022-09-13
[g06vJGtc] Fix flaky tests due to SocketTimeoutException - added assertEventually

vga91:
Since the exception is thrown when we read the url https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.4/src/test/resources/testload.zip?raw=true!xml/books.xml, and similar,
we have no control to the timeout error due to slow response.

So I replaced the `testResult` and` testCall` with `assertEventually` to prevent future `SocketTimeoutException`s, 
 e.g. in https://trello.com/c/g06vJGtc/951-essential-builds-failure-apocloadxmltesttestloadxmlfromtargzbyurl-is-failing-on-a-dev-essential-build?search_id=0b3b4048-6ce6-44fd-9469-726d051c66b5

### DIFF
--- a/core/src/test/java/apoc/load/LoadJsonTest.java
+++ b/core/src/test/java/apoc/load/LoadJsonTest.java
@@ -33,6 +33,7 @@ import static apoc.convert.ConvertJsonTest.EXPECTED_AS_PATH_LIST;
 import static apoc.convert.ConvertJsonTest.EXPECTED_PATH;
 import static apoc.convert.ConvertJsonTest.EXPECTED_PATH_WITH_NULLS;
 import static apoc.util.MapUtil.map;
+import static apoc.util.TestUtil.assertEventuallySilently;
 import static apoc.util.TestUtil.testCall;
 import static apoc.util.TestUtil.testResult;
 import static java.util.Arrays.asList;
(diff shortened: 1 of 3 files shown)


vga91:
[[Essential builds failure] apoc.load.XmlTest.testLoadXmlFromZipByUrl is failing on a Dev essential build](https://trello.com/c/FO9tIWDL/1096-essential-builds-failure-apocloadxmltesttestloadxmlfromzipbyurl-is-failing-on-a-dev-essential-build)

nadja-muller (approved):
Sounds like something which could solve the prolbem :+1:

core/src/test/java/apoc/load/LoadJsonTest.java
```diff
@@ -259,46 +260,35 @@ public void testLoadJsonWithPathOptions() throws Exception {
...
     @Test public void testLoadJsonZipByUrl() throws Exception {
         URL url = new URL("https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/3.4/src/test/resources/testload.zip?raw=true");
-        testCall(db, "CALL apoc.load.json($url)",map("url",url.toString()+"!person.json"),
-                (row) -> {
-                    Map<String,Object> r = (Map<String, Object>) row.get("value");
-                    assertEquals("Michael", r.get("name"));
-                    assertEquals(41L, r.get("age"));
-                    assertEquals(asList("Selina", "Rana", "Selma"), r.get("children"));
-                });
+        assertEventuallySilently(db, "CALL apoc.load.json($url)", map("url", url +"!person.json"),
```
ncordon, line 263: Is this going to swallow any error that happens inside the CALL command?

ncordon (changes requested):
I'm worried this is ignoring all exceptions

core/src/test/java/apoc/load/LoadJsonTest.java
vga91: I changed the `assertEventuallySilently` in order to fails only with `SocketTimeoutException` class

nadja-muller:
This test is no longer considered flaky.

PR #179, opened 2022-09-13
Fixes #25: Incorrect behaviours for format durations

### LINKED ISSUE #25: Incorrect behaviours for format durations
<a href="https://github.com/vga91"></img></a> **Issue by [vga91](https://github.com/vga91)**
_Wednesday Mar 16, 2022 at 08:59 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/2642_

----

```
WITH duration.between(  datetime("2019-06-02T19:40:32"), datetime("2019-06-03T19:40:32") ) AS duration
RETURN apoc.temporal.format(duration, "dd 'days'") AS value
```
returns `"02 days"`, and

```
WITH duration.between(  datetime("2019-06-02T19:40:32"), datetime("2019-06-02T19:40:32") ) AS duration
RETURN apoc.temporal.format(duration, "dd 'days'") AS value
```
returns `"01 days"`.

vga91:
Fixes #25

### DIFF
--- a/core/src/main/java/apoc/temporal/TemporalProcedures.java
+++ b/core/src/main/java/apoc/temporal/TemporalProcedures.java
@@ -7,9 +7,12 @@ import org.neo4j.values.storable.DurationValue;
 
 import java.time.*;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
+import java.util.Map;
 
 import static apoc.date.Date.*;
 import static apoc.util.DateFormatUtil.*;
+import static org.neo4j.values.storable.DurationFields.MONTHS;
 
 public class TemporalProcedures
 {
@@ -65,15 +68,25 @@ public class TemporalProcedures
             @Name("input") Object input,
             @Name("format") String format
     ) {
+        
+        DurationValue duration = ((DurationValue) input);
+        
+        final long months = duration.get(MONTHS.name()).value();
+        if (months > 0) {
+            duration = duration
+                    .plus(-1, ChronoUnit.DAYS)
+                    .plus(-1, ChronoUnit.MONTHS);
+        }
+        
         try {
             LocalDateTime midnight = LocalDateTime.of(0, 1, 1, 0, 0, 0, 0);
-            LocalDateTime newDuration = midnight.plus( (DurationValue) input );
+            LocalDateTime newDuration = midnight.plus(duration);
 
             DateTimeFormatter formatter = getOrCreate(format);
 
             return newDuration.format(formatter);
         } catch (Exception e){
-        throw new RuntimeException("Available formats are:\n" +
+            throw new RuntimeException("Available formats are:\n" +
                 String.join("\n", getTypes()) +
                 "\nSee also: https://www.elastic.co/guide/en/elasticsearch/reference/5.5/mapping-date-format.html#built-in-date-formats " +
                 "and https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html");
--- a/core/src/test/java/apoc/temporal/TemporalProceduresTest.java
+++ b/core/src/test/java/apoc/temporal/TemporalProceduresTest.java
@@ -64,6 +64,13 @@ public class TemporalProceduresTest
         String output = TestUtil.singleResultFirstColumn(db, "RETURN apoc.temporal.format( duration('P0M0DT4820.487660000S'), \"HH:mm:ss.SSSS\" ) as output");
         assertEquals("01:20:20.4876", output);
     }
+    
+    @Test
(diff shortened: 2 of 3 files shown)


core/src/main/java/apoc/temporal/TemporalProcedures.java
```diff
@@ -65,15 +67,25 @@ public String formatDuration(
...
+        if (months > 0) {
+            duration = duration
+                    .plus(-1, ChronoUnit.DAYS)
+                    .plus(-1, ChronoUnit.MONTHS);
+        }
+        
         try {
             LocalDateTime midnight = LocalDateTime.of(0, 1, 1, 0, 0, 0, 0);
-            LocalDateTime newDuration = midnight.plus( (DurationValue) input );
+            LocalDateTime newDuration = midnight.plus(duration);
```
AzuObs, line 82: The initial implementation of this procedure, dating back to 2018, seems to me to be a little obscure and error-prone. Since we are now aware that this piece of code is not working, may I suggest that we take a more holistic look at the problem and see if there isn't a better way to solve this problem which is simpler and more maintainable over the long term?

Cypher's [duration format](https://neo4j.com/docs/cypher-manual/current/functions/temporal/duration/) seems to be structured in a way that would make it easy to parse. 

Since all this APOC procedure does is reformat the Cypher output to a new format, may I suggest that we implement some sort of format-mapper between the two?

It seems like this would be simpler to maintain over the long term than continuing down the path of writing date arithmetic. WDYT?

core/src/test/java/apoc/temporal/TemporalProceduresTest.java
```diff
@@ -64,6 +64,13 @@ public void shouldFormatDuration() throws Throwable
...
     }
+    
+    @Test
+    public void shouldFormatDurationGreaterThanADay() {
+        String output = TestUtil.singleResultFirstColumn(db, "WITH duration.between(datetime(\"2017-06-02T18:40:32\"), datetime(\"2019-07-03T19:41:33\")) AS duration\n" +
+                "RETURN apoc.temporal.formatDuration(duration, \"yy 'years' MM 'months' dd 'days' - HH:mm:ss\") AS value");
+        assertEquals("02 years 01 months 01 days - 01:01:01", output);
+    }
 
     @Test
```
AzuObs, line 75: It would probably be nice to have more tests in the future. For example, I don't think the test below would pass.

```
    @Test
    public void test() {
        String output = TestUtil.singleResultFirstColumn(db, "WITH duration.between(datetime(\"2019-07-03T18:40:32\"), datetime(\"2019-07-03T19:41:33\")) AS duration\n" +
                "RETURN apoc.temporal.formatDuration(duration, \"yy 'years' MM 'months' dd 'days' - HH:mm:ss\") AS value");
        assertEquals("00 years 00 months 00 days - 01:01:01", output);
    }
```

AzuObs (commented):
Thank you for the hard work 🎉 

I would like to propose that this is a good opportunity to simplify the implementation of the function, but please let me know if you feel any different or if you feel like there is something I'm more understanding, and I will be happy to discuss!

core/src/main/java/apoc/temporal/TemporalProcedures.java
vga91: Not sure I really understand what you mean 😅.

You intend, extract the duration components, like `years`, `days`, etc.. , e.g. in this way:
```
 import static org.neo4j.values.storable.DurationFields.*;
 ...
final long years = duration.get(YEARS.propertyKey).value();
final long months = duration.get(MONTHS_OF_YEAR.propertyKey).value();
final long days = duration.get(DAYS.propertyKey).value();
final long hours = duration.get(HOURS.propertyKey).value();
final long minutes = duration.get(MINUTES_OF_HOUR.propertyKey).value();
final long seconds = duration.get(SECONDS_OF_MINUTE.propertyKey).value();
final long nanos = duration.get(NANOSECONDS_OF_SECOND.propertyKey).value();
```

and then apply a sort of `String.format (...)`? Because in this case we cannot use `DateTimeFormatter`s, e.g. `iso_date_time`.

Or maybe place the components in the `LocalDateTime.of(...)`? But in this case we should anyway add `1` at month and days, because it doesn't accept `0`s.

Maybe that's not what you want to implement? Sorry if you mean anything else..

vga91: @AzuObs I just changed the implementation. The duration is now handled by an ad-hoc implementation, rather than by `DateTimeFormatter`. I made it as consistent as possible with the latter, but obviously some things work differently (e.g. there are no timezone and days of the week)

AzuObs (approved):
Fantastic job on this. Thank you so much for making all the changes! 🎉

PR #186, opened 2022-09-15
[UHxIZ8JO] Add CodeQL

AzuObs:
- Ignore automated security PR from JLLeitschuh
- Install CodeQL

The CodeQL job is not decoupled from the main CI job. This is because CodeQL requires the artefacts to be built, and I didn't think it was worth to build the project twice.


common/src/main/java/apoc/export/util/ElementCounter.java
```diff
@@ -1,51 +0,0 @@
-package apoc.export.util;
-
-/**
- * @author mh
- * @since 17.01.14
- */
-public class ElementCounter {
```
ncordon, line 7: Is this used in extended?

ncordon (commented):
LGTM with the exception of the small comment I left

ncordon (approved):

common/src/main/java/apoc/export/util/ElementCounter.java
ncordon: Daniel has confirmed this was not used in extended

PR #187, opened 2022-09-16
[VP7iRcuF] Introduce helper methods from neo4j

Lojjs:
In order to decrease the dependencies of internal Neo4j APIs, it was decided to copy over some of the helper classes into APOC.

A brief list of proposed changes in order to fix the issue:

  - Copy over all relevant parts of the org.neo4j.internal.helper.collection package from Neo4j
  - Remove benchmarking code which is old and unused

## Notes for the reviewer

- Some of the helper files which have been copied over from Neo4j also had accompanying test classes. Do we want to move over relevant parts of these as well?
- This will break APOC extended, as several places there call public core methods which have now changed signature from e.g. org.neo4j.internal.helpers.collection.Pair to apoc.util.collection.Pair. Should we make a Trello card for fixing this?


common/src/main/java/apoc/convert/ConvertUtils.java
```diff
@@ -1,7 +1,6 @@
 package apoc.convert;
 
-import org.neo4j.internal.helpers.collection.Iterators;
-
+import apoc.util.collection.Iterators;
```
vga91, line 3: Not directly related, but we could change rows 20 and 21 with these?
```
        else if (list instanceof Iterable) return Iterables.asList((Iterable)list);
        else if (list instanceof Iterator) return Iterators.asList((Iterator)list);
```

common/src/main/java/apoc/util/collection/Iterators.java
```diff
@@ -0,0 +1,225 @@
...
+     * contract of {@link Iterable} in that it returns the supplied iterator
+     * instance for each call to {@code iterator()} on the returned
+     * {@link Iterable} instance. This method exists to make it easy to use an
+     * {@link Iterator} in a for-loop.
+     *
+     * @param <T> the type of items in the iterator.
+     * @param iterator the iterator to expose as an {@link Iterable}.
+     * @return the supplied iterator posing as an {@link Iterable}.
+     */
+    public static <T> Iterable<T> loop(final Iterator<T> iterator) {
```
vga91, line 139: I think we could also use `Iterator.forEachRemaining` instead of this?

vga91 (commented):
A really good job :), I just added some small (mostly subjective) things I think could be changed

vga91:
> * Some of the helper files which have been copied over from Neo4j also had accompanying test classes. Do we want to move over relevant parts of these as well?
> * This will break APOC extended, as several places there call public core methods which have now changed signature from e.g. org.neo4j.internal.helpers.collection.Pair to apoc.util.collection.Pair. Should we make a Trello card for fixing this?

For me yes, both, I think it is worth adding the some tests and the `APOC extended` trello card

common/src/main/java/apoc/convert/ConvertUtils.java
Lojjs: Nice catch!

common/src/main/java/apoc/util/collection/Pair.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Objects;
+
+/**
+ * Utility to handle pairs of objects.
+ *
+ * @param <T1> the type of the {@link #first() first value} of the pair.
+ * @param <T2> the type of the {@link #other() other value} of the pair.
+ */
+public abstract class Pair<T1, T2> {
```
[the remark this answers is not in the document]
Lojjs: Hmm I guess we could, as we already depend on org.apache.commons.lang3

common/src/test/java/apoc/util/collection/AbstractResourceIterableTest.java
```diff
@@ -0,0 +1,320 @@
...
+            @Override
+            protected void onClosed() {
+                iterableClosed.setTrue();
+            }
+        };
+        final var iterator = iterable.iterator();
+
+        // Then
+        assertThat( Iterators.asList(iterator)).containsExactlyElementsOf(items);
+        assertThat(iteratorClosed.isTrue()).isTrue();
```
vga91, line 54: This seems a bit repetitive, at least for me. Maybe I would write:
```suggestion
        assertTrue(iteratorClosed.booleanValue());
```
or 
```suggestion
        assertTrue(iteratorClosed.getValue());
```

Same thing for other similar ones (lines 55, 86, 114-117, 167, 194-196, 232-234, 264-266, 295,296, 318).

vga91 (approved):
I'd just change one more thing (optional). Except this one LGTM :)

common/src/test/java/apoc/util/collection/AbstractResourceIterableTest.java
Lojjs: I ended up doing the first suggestion and applied it for the other test files as well

PR #190, opened 2022-09-21
Fixes #18: Default parallel concurrency of 50 is too high

### LINKED ISSUE #18: Default parallel concurrency of 50 is too high
<a href="https://github.com/wdroste"></img></a> **Issue by [wdroste](https://github.com/wdroste)**
_Wednesday Jun 08, 2022 at 18:27 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/2971_

----

For parallel often times there's only 4 CPU or if there's 50 CPU its just too many threads. I would expect a default of 4.
apoc.iterate('delete query', { batchSize: 1000, parallel: true})

I expected this to be 4-6

The actual behavior is 50 threads, which either tends to lock on deletes or over tax the system

Run a delete on 1m nodes on a small box.

```
apoc.iterate("MATCH (n)", "DETACH DELETE n", { batchSize: 1000, parallel: true})
```

### Versions

  - OS: NA
  - Neo4j: 4.0+
  - Neo4j-Apoc: 4.0+

vga91:
Fixes #18

Changed 50 with number of processors available

### DIFF
--- a/core/src/main/java/apoc/periodic/Periodic.java
+++ b/core/src/main/java/apoc/periodic/Periodic.java
@@ -239,7 +239,7 @@ public class Periodic {
      * @param cypherAction
      */
     @Procedure(mode = Mode.WRITE)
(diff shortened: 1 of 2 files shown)


gem-neo4j (approved):
LGTM :)

PR #200, opened 2022-09-26
Fixes #11: apoc.load* don't actually support tar and tar.gz

### LINKED ISSUE #11: apoc.load* don't actually support tar and tar.gz
<a href="https://github.com/nitay-bachrach"></img></a> **Issue by [nitay-bachrach](https://github.com/nitay-bachrach)**
_Thursday Aug 25, 2022 at 17:53 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/3151_

----

According to https://neo4j.com/labs/apoc/4.3/import/web-apis/ and other sources, apoc.load.json, apoc.load.csv etc should support .tar, .tar.gz, and .tgz

CALL apoc.load.json("https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/4.3/core/src/test/resources/testload.tgz?raw=true!person.json");

In practice, only .zip is supported
In fact, all the test files here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/tree/4.4/full/src/test/resources
and here:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/tree/4.4/core/src/test/resources

are just zip fies with different extensions. They all share the same md5.

Create an actual tar/tar.gz/tgz file (using `tar -xf` or `tar -xzf`) and  try to load it
### Steps (Mandatory)

  1. echo {"a":"b"} > test.json
  2. . tar -czf test.tar.gz test.json
  3. python -m http.server 8080
  4. CALL apoc.load.json("http://localhost:8080/test.tar.gz!test.json")

## Specifications (Mandatory)

All versions

### Versions

  - OS: Windows/Mac/Linux
  - Neo4j: 4.*
  - Neo4j-Apoc: *

vga91:
Fixes #11

Tests modified with "true" tar and tgz (instead of `zip` renamed to` tar` / `tgz` /` tar.gz`)

Added `ArchiveType` factory which recognizes the file from the extension


gem-neo4j:
Looks good :) Just need to fix up the unused imports as it seems to be failing the spotless check

gem-neo4j (approved):

PR #202, opened 2022-09-29
[hcaaOlXR] Various security fixes

AzuObs:
Makes XML procedures resistant to XXE attacks by following the general advice which is to disable External Entities. This is what we had done previously for apoc.load.xml and apoc.xml.parse, but forgot to do for apoc.import.xml.


gem-neo4j (approved):
Looks good! :D

PR #205, opened 2022-10-05
[hcaaOlXR] Protect from Billion Laughs

AzuObs:
Protects fully against XXE and Billion Laugh attacks by following the following recommendations:

- [Owasp recommendation](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#xmlinputfactory-a-stax-parser) 
- [Leftshift recommendation](https://www.shiftleft.io/blog/preventing-xxe-in-java-applications-02-12-2021/#xmlinputfactory)

### DIFF
--- a/core/src/main/java/apoc/load/Xml.java
+++ b/core/src/main/java/apoc/load/Xml.java
@@ -26,6 +26,7 @@ import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.InputSource;
+import org.xml.sax.SAXParseException;
 
 import javax.xml.namespace.QName;
 import javax.xml.parsers.DocumentBuilder;
@@ -69,6 +70,7 @@ public class Xml {
     private static final XMLInputFactory FACTORY = XMLInputFactory.newFactory();
     static {
         FACTORY.setProperty(XMLInputFactory.IS_COALESCING, true);
+        FACTORY.setProperty(XMLInputFactory.SUPPORT_DTD, false);
         FACTORY.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);
     }
 
@@ -148,6 +150,8 @@ public class Xml {
         catch (Exception e){
             if(!failOnError)
                 return Stream.of(new MapResult(Collections.emptyMap()));
+            else if (e instanceof SAXParseException && e.getMessage().contains("DOCTYPE is disallowed"))
+                throw generateXmlDoctypeException();
             else
                 throw e;
         }
@@ -501,6 +505,9 @@ public class Xml {
             xml.next();
 
             switch (xml.getEventType()) {
+                case XMLStreamConstants.DTD:
+                    throw generateXmlDoctypeException();
+
                 case XMLStreamConstants.START_DOCUMENT:
                     // xmlsteamreader starts off by definition at START_DOCUMENT prior to call next() - so ignore this one
                     break;
@@ -604,4 +611,7 @@ public class Xml {
         }
     }
 
+    private RuntimeException generateXmlDoctypeException() {
+        throw new RuntimeException("XML documents with a DOCTYPE are not allowed.");
+    }
 }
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/load/XmlTest.java
```diff
@@ -466,6 +466,21 @@ public void testLoadXmlPreventXXEVulnerabilityThrowsQueryExecutionException() {
...
+    public void testLoadXmlPreventBillionLaughVulnerabilityThrowsQueryExecutionException() {
+        try {
+            testResult(db, "CALL apoc.load.xml('" + TestUtil.getUrlFileName("xml/billion_laughs.xml") + "')", (r) -> {
+                r.next();
+                r.close();
+            });
+        } catch (Exception e) {
+            Throwable except = ExceptionUtils.getRootCause(e);
+            assertTrue(except instanceof SAXParseException);
+            assertEquals("DOCTYPE is disallowed when the feature \"http://apache.org/xml/features/disallow-doctype-decl\" set to true.", except.getMessage());
```
ncordon, line 479: Should we be surfacing errors in the underlying library?

ncordon (commented):
LGTM, just a small comment, but I'm happy to merge as is

ncordon (approved):

core/src/main/java/apoc/load/Xml.java
```diff
@@ -69,6 +69,7 @@ public class Xml {
     private static final XMLInputFactory FACTORY = XMLInputFactory.newFactory();
     static {
         FACTORY.setProperty(XMLInputFactory.IS_COALESCING, true);
+        FACTORY.setProperty(XMLInputFactory.SUPPORT_DTD, false);
         FACTORY.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);
```
DimitrisLarisis, line 73: not a Java expert, but what is the difference between this implementation and what OWASP snippet looks like: `xmlInputFactory.setProperty("javax.xml.stream.isSupportingExternalEntities", false);`

DimitrisLarisis: also , what about this one:
`// This causes XMLStreamException to be thrown if external DTDs are accessed.
xmlInputFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");`.
I guess it is not needed , because right above we set `SUPPORT_DTD` to false?

AzuObs: >not a Java expert, but what is the difference between this implementation and what OWASP snippet looks like: xmlInputFactory.setProperty("javax.xml.stream.isSupportingExternalEntities", false);

`FACTORY.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);` is the same as the line above. It's just using the Enum object rather than writing it out as a String.

> also , what about this one:
> // This causes XMLStreamException to be thrown if external DTDs are accessed. 
> xmlInputFactory.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");.
> I guess it is not needed , because right above we set SUPPORT_DTD to false?

I think the OWASP article was for an older version of Java perhaps, where the StAX parser returned was using this configuration. The default parser Java is giving us does not use the setting. 

The setting is not entirely needed however as it simply throws an exception whenever the parser encounters a `<!DOCTYPE>`. Currently what our implementation is doing is that it simply ignores the `<!DOCTYPE>` and will throw an error if it encounters an unknown entity (eg one that was declared the in DOCTYPE)

This is not a bad thing really as it makes my change a bit more backwards compatible while still being safe.

It's also perfectly aligned with the LeftShift (eg competitor to Snyk and Veracode) recommendation which is to implement exactly what I've done. It's mostly aligned with the OWASP article but only because I suspect the OWASP article is quite old and doesn't fully work anymore.

DimitrisLarisis: Thanks for looking into this!! Let's merge :)

DimitrisLarisis (approved):

PR #207, opened 2022-10-07
[9uFWHlVy] Check compatibility on system database instead of default database

Lojjs:
## Fixes
Startup log exception in Neo4j 5.x about 'Topology Graph is not initialized yet' 

  - Check compatibility on the system database instead of the default database
  - Make sure dbms.components() has been registered 
  - Update compatibility warning message to apply to 5.x versioning

### DIFF
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -10,7 +10,6 @@ import org.neo4j.common.DependencyResolver;
 import org.neo4j.configuration.GraphDatabaseSettings;
 import org.neo4j.kernel.api.procedure.GlobalProcedures;
 import org.neo4j.kernel.availability.AvailabilityListener;
-import org.neo4j.kernel.database.DefaultDatabaseResolver;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.logging.Log;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -62,16 +61,16 @@ public void available() {
...
                         final List<String> versions = db.executeTransactionally("CALL dbms.components", Collections.emptyMap(),
                                 r -> (List<String>) r.next().get("versions"));
-                        final String apocFullVersion = Version.class.getPackage().getImplementationVersion();
-                        if (isVersionDifferent(versions, apocFullVersion)) {
+                        final String apocVersion = Version.class.getPackage().getImplementationVersion();
+                        if (isVersionDifferent(versions, apocVersion)) {
                             userLog.warn("The apoc version (%s) and the Neo4j DBMS versions %s are incompatible. \n" +
-                                            "See the compatibility matrix in https://neo4j.com/labs/apoc/4.4/installation/ to see the correct version",
-                                    apocFullVersion, versions.toString());
+                                            "The two first numbers of both versions needs to be the same.",
```
gem-neo4j, line 72: Is this true? I thought after 5.0 we wanted to have a compatibility layer that works differently

Lojjs: Me, Nacho and Daniel discussed this a bit at the lunch lecture on GDS compatibility layer last week. It is leaning towards that we will not do a compatibility layer, but just say you need to use  e.g. APOC 5.2.0 with Neo4j 5.2.0 or Neo4j 5.2.1 (if there is a security patch)

gem-neo4j: okay :)

gem-neo4j (approved):
Looks good to me, the only thing I can think of that might be useful is to add a comment to explain why this works as I was confused :P But that is up to you :)

PR #208, opened 2022-10-11
Fixes neo4j/apoc#126: apoc.periodic.submit fails with schema operations (core)

vga91:
Fixes neo4j/apoc#126 (core part)

1) Added QueryType validations in `validateQuery()` method, so as to fail if the procedure tries to execute schema operations

2) Refactored `apoc.periodic.submit` code in order to be reused in extended.

3) In  https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3211, added `apoc.periodic.submitSchema` procedure, equivalent but with different `validateQuery()` and `org.neo4j.procedure.Mode.SCHEMA`

----

NOTE FOR REVIEWER:
- We could also create a schema equivalent for the others periodic procedures (maybe in another issue/pr?)
- Even the `apoc.trigger.add` has the `validateQuery()`, maybe can be changed in the same way? (also in this case, perhaps in another pr)

### DIFF
--- a/core/src/test/java/apoc/periodic/PeriodicTest.java
+++ b/core/src/test/java/apoc/periodic/PeriodicTest.java
@@ -25,7 +25,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -79,6 +79,18 @@ public void testSubmitStatement() throws Exception {
         testCall(db, callList, (r) -> assertEquals(true, r.get("done")));
     }
 
+    @Test
+    public void testSubmitWithSchemaOperation() {
+        try {
+            testCall(db, "CALL apoc.periodic.submit('subSchema','CREATE INDEX periodicIdx FOR (n:Bar) ON (n.first_name, n.last_name)')",
```
Lojjs, line 85: When you cherry-pick to 4.x, it would be good to also add a test for `CALL db.createUniquePropertyConstraint` (that has been removed in Neo4j 5.0) to make sure this fix the original reported issue.

PR #214, opened 2022-10-18
[hDJkPkOQ] Add QueryType validation in apoc.trigger.add

vga91:
TODO - Waiting for https://github.com/neo4j/apoc/pull/250

Added QueryType validations in `validateQuery()` method, 
so as to fail if the procedure tries to execute schema operations.

See point 2 here: https://github.com/neo4j/apoc/pull/208#pullrequestreview-1145222227

### DIFF
--- a/core/src/main/java/apoc/trigger/Trigger.java
+++ b/core/src/main/java/apoc/trigger/Trigger.java
@@ -12,6 +12,10 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.stream.Stream;
 
+import static org.neo4j.graphdb.QueryExecutionType.QueryType.READ_ONLY;
+import static org.neo4j.graphdb.QueryExecutionType.QueryType.READ_WRITE;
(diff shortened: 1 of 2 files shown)


vga91:
> TODO - Waiting for #250
> 
> Added QueryType validations in `validateQuery()` method, so as to fail if the procedure tries to execute schema operations.
> 
> See point 2 here: [#208 (review)](https://github.com/neo4j/apoc/pull/208#pullrequestreview-1145222227)

I'm afraid I waited in vain :X
I think this pr is no longer valid on 5.x, as in `apoc.trigger.install` we removed the `Util.validateQuery(db, statement)`  ( [see here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183#discussion_r1005547491) ),
while the `apoc.trigger.add` is now deprecated.

So, I just created a pr for 4.4 https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3421

PR #216, opened 2022-10-18
Update testJaroWinklerDistance test expectation

MishaDemianenko:
Before this PR test was expecting the incorrect value of distance.
As a result, https://issues.apache.org/jira/browse/TEXT-191 and the new commons-text test
is updated with the correct expected value.


AzuObs (approved):
LGTM 🎉

PR #218, opened 2022-10-19
Fixes #126: Improve validateQuery

### LINKED ISSUE #126: can not get apoc.periodic.submit() to run CALL db.createUniquePropertyConstraint()
<a href="https://github.com/Dcanzano"></img></a> **Issue by [Dcanzano](https://github.com/Dcanzano)**
_Friday Feb 08, 2019 at 14:27 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1089_

----

## Guidelines

Please note that GitHub issues are only meant for bug reports/feature requests. If you have questions on how to use Streams, please ask on [StackOverflow](https://stackoverflow.com/questions/tagged/neo4j-apoc) instead of creating an issue here.

Unable to get apoc.periodic.submit() to call `CALL db.createUniquePropertyConstraint(":E(id)", "lucene-1.0")`  as the constraint and associated index are never created.  No errors, nothing written to debug.log to describe failure

# Actual Behavior (Mandatory)
expect the constraint and related index to be created

have tried
call apoc.periodic.submit('job123','CALL db.createUniquePropertyConstraint(":E(id)","lucene-1.0");');

or

call apoc.periodic.submit('job123','CALL db.createUniquePropertyConstraint(\":E(id)\",\"lucene-1.0\");');

but still no success

### Simple Dataset (where it's possibile)

```
//Insert here a set of Cypher statements that helps us to reproduce the problem
see above
```

### Steps (Mandatory)
see above

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS:
 uname -a
Linux ubuntu17 4.13.0-16-generic #19-Ubuntu SMP Wed Oct 11 18:33:49 UTC 2017 i686 i686 i686 GNU/Linux

  - Neo4j:   3.5.2
  - Neo4j-Apoc:   apoc-3.5.0.1-all.jar

vga91:
After this cherry-pick https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3234,
I realized that the validateQuery doesn't work with an inner `org.neo4j.procedure.Mode.SCHEMA` procedure, e.g. `CALL db.createUniquePropertyConstraint`,
so I improved it to check these situations.

The `validateQuery` still has some limitations, for example it does not catch cases like this `CALL apoc.cypher.run (\" CALL apoc.schema.assert ({}, {}) \ ", {}) YIELD value RETURN value`, with a sub-inner schema procedure,

or I might have a `call apoc.cypher.runSchema ('return 1', {})` which even though it's a `Mode.SCHEMA`, doesn't actually perform schema operations (even if it doesn't make much sense).

Perhaps, in some way, the above 2 cases can also be validated,
but I think they could be very complex to verify and I don't know if it's worthwhile.

### DIFF
--- a/core/src/main/java/apoc/periodic/Periodic.java
+++ b/core/src/main/java/apoc/periodic/Periodic.java
@@ -22,9 +22,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
-import static org.neo4j.graphdb.QueryExecutionType.QueryType.READ_ONLY;
-import static org.neo4j.graphdb.QueryExecutionType.QueryType.READ_WRITE;
-import static org.neo4j.graphdb.QueryExecutionType.QueryType.WRITE;
+import static org.neo4j.graphdb.QueryExecutionType.QueryType;
 import static apoc.periodic.PeriodicUtils.recordError;
 import static apoc.periodic.PeriodicUtils.submitJob;
 import static apoc.periodic.PeriodicUtils.submitProc;
@@ -177,8 +175,9 @@ public class Periodic {
     }
 
     private void validateQuery(String statement) {
-        Util.validateQuery(db, statement,
-                READ_ONLY, WRITE, READ_WRITE);
+        Util.validateQuery(db, statement, 
+                Set.of(Mode.WRITE, Mode.READ, Mode.DEFAULT),
+                QueryType.READ_ONLY, QueryType.WRITE, QueryType.READ_WRITE);
     }
 
     @Procedure(mode = Mode.WRITE)
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -62,6 +62,7 @@ import java.util.Scanner;
 import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
+import java.util.TreeSet;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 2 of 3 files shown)


core/src/main/java/apoc/periodic/Periodic.java
```diff
@@ -177,8 +175,9 @@ public Stream<JobInfo> repeat(@Name("name") String name, @Name("statement") Stri
     }
 
     private void validateQuery(String statement) {
-        Util.validateQuery(db, statement,
-                READ_ONLY, WRITE, READ_WRITE);
+        Util.validateQuery(db, statement, 
+                Set.of(Mode.WRITE, Mode.READ, Mode.DEFAULT),
+                QueryType.READ_ONLY, QueryType.WRITE, QueryType.READ_WRITE);
```
Lojjs, line 180: Now we will fail with `Supported query types for the operation are [READ_ONLY, WRITE, READ_WRITE]` regardless if it is the mode or querytype that is violated, do we want to write it so we fail with different error messages instead?

Lojjs (changes requested):
Looks good in general, I just have one comment plus a super small one.

Your example from the description: `CALL apoc.cypher.run (\" CALL apoc.schema.assert ({}, {}) \ ", {}) YIELD value RETURN value` will fail anyway as you cannot run a schema procedure inside a non-schema procedure. Also `call apoc.cypher.runSchema ('return 1', {})` I think is ok to invalidate even if it doesn't perform any schema, I would say it is more of a user issue. So I don't think we need to make it more powerful.

core/src/main/java/apoc/periodic/Periodic.java
vga91: Yes  it makes sense, I changed the `validateQuery` method in order to distinguish the two cases

Lojjs (approved):

PR #219, opened 2022-10-20
[4HrPR2Rm] Use internal function instead of procedure for compatibility check.

Lojjs:
This has been requested by Neo4j Aura, which couldn't see if the procedure call came from APOC or a user

### DIFF
--- a/core/src/test/java/apoc/cypher/CypherIsVersionDifferentTest.java
+++ b/core/src/test/java/apoc/cypher/CypherIsVersionDifferentTest.java
@@ -17,6 +17,9 @@ public class CypherIsVersionDifferentTest {
         assertFalse(isVersionDifferent("4.4.5", "4.4.0.4"));
         assertFalse(isVersionDifferent("5.1.1", "5.1.0"));
         assertTrue(isVersionDifferent("5.1.0", "5.2.0"));
+        assertFalse(isVersionDifferent("5.1-aura", "5.1.0"));
+        assertTrue(isVersionDifferent("4.4-aura", "5.1.0"));
+        assertTrue(isVersionDifferent("5.2-aura", "5.1.0"));
 
         // we expect that APOC versioning is always consistent to Neo4j versioning
         assertTrue(isVersionDifferent("", "5_2_0_1"));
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/cypher/CypherIsVersionDifferentTest.java
```diff
@@ -12,30 +10,33 @@ public class CypherIsVersionDifferentTest {
...
-        assertTrue(isVersionDifferent(List.of("5.1.0"), "5.2.0"));
+        assertTrue(isVersionDifferent("3.5", "4.4.0.2"));
+        assertTrue(isVersionDifferent("5_0", "4.4.0.2"));
+        assertFalse(isVersionDifferent("3.5.12", "3.5.0.9"));
+        assertFalse(isVersionDifferent("3.5.12", "3.5.1.9"));
+        assertFalse(isVersionDifferent("4.4.5", "4.4.0.4"));
+        assertFalse(isVersionDifferent("5.1.1", "5.1.0"));
+        assertTrue(isVersionDifferent("5.1.0", "5.2.0"));
+        assertFalse(isVersionDifferent("5.1-aura", "5.1.0"));
+        assertTrue(isVersionDifferent("4.4-aura", "5.1.0"));
```
nadja-muller, line 21: Is that correct? :sweat_smile:

Lojjs: Yes I think so. The method checks if the versions are **not** compatible. So this checks that Neo4j 5.1.x and 5.1-aura are compatible with apoc 5.1.x while the combos Neo4j 5.1.x/APOC 5.2.x resp. Neo4j 4.4-aura/APOC 5.1.x are not compatible

nadja-muller: I completely missed the assertTrue/asertFalse difference. Thank you for the explanation :+1:

nadja-muller (approved):
Looks like there was a flaky test failure in the CI, but otherwise it LGTM :smile:

PR #222, opened 2022-10-23
[vgniAKNX] Fixes #36: Allow to treat JSON numbers as strings

### LINKED ISSUE #36: Allow to treat JSON numbers as strings
<a href="https://github.com/odessit55"></img></a> **Issue by [odessit55](https://github.com/odessit55)**
_Tuesday Jul 20, 2021 at 19:18 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/2109_

----

When processing following JSON file:
{"info_file_properties_modified_time":"18446744062065078016",
"info_file_properties_modified_time2":18446744062065078016}
using 
apoc.load.json() 
We get following error:
Failed to invoke procedure `apoc.load.json`: Caused by: com.fasterxml.jackson.core.exc.InputCoercionException: Numeric value (18446744062065078016) out of range of long (-9223372036854775808 - 9223372036854775807)
 at [Source: (apoc.export.util.CountingInputStream); line: 3, column: 59]

Following JSON works fine as I removed a trailing "6" from the string
{"info_file_properties_modified_time":"18446744062065078016",
"info_file_properties_modified_time2":1844674406206507801}

It would be great if we can specify that certain JSON fields should be treated as strings/larger numbers or at least to treat all strings as numbers.

Thank you!

vga91:
Fixes #36 

Added `CustomNumberDeserializer`, settable via `allowBigNum` config to coerce Numbers > `Long.MAX_VALUE` or `Double.MAX_VALUE` in String

---

Docs here: https://github.com/neo4j/docs-apoc/pull/54


gem-neo4j:
Hey! Thanks for doing this, I've talked with the team and we don't think this extension should be added to the procedure.

Our thoughts are that long is the largest Java Type we support for Cypher Integers and Double for Cypher Floats. So the exception is probably a good thing.

And having only numbers that would error get turned into strings creates a weird consistency issue.

We basically don't want to introduce something we won't want to port to Cypher later on.

Sorry you spent time on this, but we think this and the github issue should be closed explaining this.

PR #238, opened 2022-11-14
[4Bmcyc2U] Try fixing flaky testDeadlockException

vga91:
Try fixing flaky testDeadlockException

Added more `SIMILAR_TO` relationships to significantly increase the probability of deadlock.

Indeed, there may be a very small probability that the test will fail, 
but I don't think there are any other solutions (apart from introducing a retry mechanism)

### DIFF
--- a/core/src/test/java/apoc/refactor/rename/RenameTest.java
+++ b/core/src/test/java/apoc/refactor/rename/RenameTest.java
@@ -2,7 +2,6 @@ package apoc.refactor.rename;
 
 import apoc.util.TestUtil;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


Lojjs (changes requested):
I tried to run the test locally about 500 times, and it didn't fail at all, but looking at the build log it looks like it failed on the Github PR run 😿 So I think we maybe need to go with another solution. 

One relevant question is what the purpose of the test is? Should we just test that any failure gives an error message for apoc.refactor.rename.type, then we could maybe try with a query that fails in a different way? Is it expected that `apoc.refactor.rename.type` should fail with a deadlock or is that actually a bug that should be fixed in that procedure? etc. If that helps, this test was introduced here https://github.com/conker84/neo4j-apoc-procedures/commit/e945e6c31fd72691e7b8c75ca4764bd6e014a2db

Lojjs (approved):
This looks more promising to me

vga91:
> This looks more promising to me

Ok thanks. Btw sorry, I forgot to answer the above answer, wanted to wait to check if it worked on the CI.
The purpose of the test is not only to verify that in case of errors it gives an "errorMessages", but to replicate a lock error like [in this issue](https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1387).

PR #239, opened 2022-11-15
[4Bmcyc2U] Fix ignored testGeocodeGoogle

vga91:
Currently the tests, except for the `"osm"` one, are quite unusable.

In fact, the `"opencage"` tests are actually ignored due to [this assumption](https://github.com/neo4j/apoc/compare/fix_ignored_testGeocodeGoogle?expand=1#diff-252f4d56949dc064efccfe8ec2a9f918c432ad1b2a224973c2fc2b9e499e90d1L84), 
so in order to run them, we should manually writ your own key instead of `<YOUR_API_KEY>`.

While the reverse `"google"` test is a false positive, as it returns (with an invalid/absent ket) an empty result and [here](https://github.com/neo4j/apoc/compare/fix_ignored_testGeocodeGoogle?expand=1#diff-252f4d56949dc064efccfe8ec2a9f918c432ad1b2a224973c2fc2b9e499e90d1L137) the return result works even if it shouldn't. 
There should be a `assertTrue(row.hasNext());` before the foreach.

So to make these tests valid, I think the possible ways are:
- delete these tests as we already have some mock tests (`SpatialTest.java`).

- plan insertion in the CIs of some key, e.g. `OPENCAGE_KEY` and `GOOGLE_KEY` and write the tests like this, for example:
```
private static final String GMAPS_KEY = System.getenv("GMAPS_KEY");

// in @Before/@BeforeClass method
apocConfig().setProperty("apoc.spatial.geocode.google.key", GMAPS_KEY);

    @Test
    public void testGeocodeGoogle() throws Exception {
          Assume.assumeNotNull(GMAPS_KEY);
...

```

### DIFF
--- a/core/src/test/java/apoc/spatial/GeocodeTest.java
+++ b/core/src/test/java/apoc/spatial/GeocodeTest.java
@@ -25,19 +25,7 @@ public class GeocodeTest {
 
     @Before
     public void initDb() throws Exception {
-        assumeRunningInCI();
-        apocConfig().setProperty("apoc.spatial.geocode.provider", "opencage");
-        apocConfig().setProperty("apoc.spatial.geocode.opencage.key", "<YOUR_API_KEY>");
-        apocConfig().setProperty("apoc.spatial.geocode.opencage.url", "https://api.opencagedata.com/geocode/v1/json?q=PLACE&key=KEY");
-        apocConfig().setProperty("apoc.spatial.geocode.opencage.reverse.url", "https://api.opencagedata.com/geocode/v1/json?q=LAT+LNG&key=KEY");
-
-//        List<String> strings = Iterators.asList(apocConfig().getConfig().getKeys("apoc.spatial.geocode.opencage"));
         TestUtil.registerProcedure(db, Geocode.class);
-    }
-
-    @Before
-    public void setUp() throws Exception {
-        assumeRunningInCI();
     }
 
     // -- with config map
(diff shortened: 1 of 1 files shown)


core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -106,10 +64,6 @@ private void testGeocodeWithThrottling(String supplier, Boolean reverseGeocode,
         assertTrue("Fast " + supplier + " took " + fast + "ms and slow took " + slow + "ms, but expected slow to be at least twice as long", (1.0 * slow / fast) > 1.2);
     }
 
-    private long testGeocode(String provider, long throttle, boolean reverseGeocode) throws Exception {
-        return testGeocode(provider, throttle, reverseGeocode, Collections.emptyMap());
-    }
-
     private long testGeocode(String provider, long throttle, boolean reverseGeocode, Map<String, Object> config) throws Exception {
         setupSupplier(provider, throttle);
 //        testConfig(provider);
```
gem-neo4j, line 69: Nothing to do with the PR, but can you remove this comment as well? it looks ugly :P

gem-neo4j (approved):

vga91:
Ok thanks. I just created a card on this.

> I think it makes sense to remove these at this time, I imagine we will have more issues with adding the Key etc, so am happy to approve this. Although can you make a low prio ticket on our board to add better test coverage for this at some time?

PR #240, opened 2022-11-16
[eOxJMLMv] Checks redirects for blocked IPs instead of just final location.

gem-neo4j:
Also fixes ignored test for URL to file redirect.

Co-authored-by: louise <louise.soderstrom@neo4j.com>

### DIFF
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -105,6 +105,7 @@ import static apoc.export.cypher.formatter.CypherFormatterUtils.formatProperties
 import static apoc.export.cypher.formatter.CypherFormatterUtils.formatToString;
 import static apoc.util.DateFormatUtil.getOrCreate;
 import static java.lang.String.format;
+import static java.net.HttpURLConnection.HTTP_NOT_MODIFIED;
 import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME;
 import static org.eclipse.jetty.util.URIUtil.encodePath;
 
@@ -118,6 +119,8 @@ public class Util {
     public static final String REL_COUNT = "MATCH ()-->() RETURN count(*) as result";
     public static final String COMPILED = "interpreted"; // todo handle enterprise properly
     public static final String ERROR_BYTES_OR_STRING = "Only byte[] or url String allowed";
+
+    public static final int REDIRECT_LIMIT = 10;
 
     public static String labelString(List<String> labelNames) {
         return labelNames.stream().map(Util::quote).collect(Collectors.joining(":"));
@@ -354,8 +357,8 @@ public class Util {
     }
 
     public static boolean isRedirect(HttpURLConnection con) throws IOException {
-        int code = con.getResponseCode();
-        boolean isRedirectCode = code >= 300 && code < 400;
+        int responseCode = con.getResponseCode();
+        boolean isRedirectCode = responseCode >= 300 && responseCode <= 307 && responseCode != 306 && responseCode != HTTP_NOT_MODIFIED;
         if (isRedirectCode) {
             URL location = new URL(con.getHeaderField("Location"));
             String oldProtocol = con.getURL().getProtocol();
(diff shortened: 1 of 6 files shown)


VariableExp0rt:
👍 from the security team

common/src/test/java/apoc/util/UtilIT.java
```diff
@@ -63,10 +57,26 @@ public void redirectShouldWorkWhenProtocolNotChangesWithUrlLocation() throws IOE
...
+        Config neo4jConfig = mock(Config.class);
+        when(neo4jConfig.get(GraphDatabaseInternalSettings.cypher_ip_blocklist)).thenReturn(blockedIPs);
+
+        setUpServer(neo4jConfig, "http://127.168.0.1");
+        String url = getServerUrl();
+
+        IOException e = Assert.assertThrows(IOException.class,
+                () -> Util.openInputStream(url, null, null, null)
+        );
+        TestCase.assertTrue(e.getMessage().contains("access to /127.168.0.1 is blocked via the configuration property internal.dbms.cypher_ip_blocklist"));
```
nadja-muller, line 73: Hm...what does the slash here signify?

common/src/test/java/apoc/util/UtilIT.java
```diff
@@ -63,10 +57,26 @@ public void redirectShouldWorkWhenProtocolNotChangesWithUrlLocation() throws IOE
         assertTrue(page.contains("<title>Google</title>"));
     }
 
-    @Ignore
+    @Test
+    public void redirectWithBlockedIPsWithUrlLocation() {
```
nadja-muller, line 61: Is this url actually redirecting?

common/src/main/java/apoc/util/Util.java
```diff
@@ -335,17 +335,19 @@ public static URLConnection openUrlConnection(String url, Map<String, Object> he
...
         con.setRequestProperty("User-Agent", "APOC Procedures for Neo4j");
-        if (headers != null) {
-            Object method = headers.get("method");
-            if (method != null && con instanceof HttpURLConnection) {
+        if (con instanceof HttpURLConnection) {
                 HttpURLConnection http = (HttpURLConnection) con;
-                http.setRequestMethod(method.toString());
-                http.setChunkedStreamingMode(1024*1024);
-                http.setInstanceFollowRedirects(true);
+                http.setInstanceFollowRedirects(false);
```
nadja-muller, line 340: When is the redirection resolved?

nadja-muller (commented):
It looks generally good to me and there is some nice cleanup in there, too :heart_eyes: 
I wonder if I miss something from the description of the PR, so I left a couple of questions.

common/src/test/java/apoc/util/UtilIT.java
gem-neo4j: This is the InetAddress from Java, it isn't added by us, it is what the host URL got resolved as in a toString. It is because it would print out hostName/address, but in this particular case there is no hostname e.g www.google.com/64.34.119.12 is possible. I think this is useful for IPs that have hostnames as a user may understand the text part more. e.g www.evilattacker.com/64.34.119.12 is easier to read than 64.34.119.12

common/src/main/java/apoc/util/Util.java
gem-neo4j: openUrlConnection is called recursively by readHttpInputStream.

```
public static StreamConnection readHttpInputStream(String urlAddress, Map<String, Object> headers, String payload) throws IOException {
        ApocConfig.apocConfig().checkReadAllowed(urlAddress);
        URLConnection con = openUrlConnection(urlAddress, headers); // CALLED HERE
        writePayload(con, payload);
        String newUrl = handleRedirect(con, urlAddress);
        if (newUrl != null && !urlAddress.equals(newUrl)) { // IF REDIRECT WILL MAKE IT IN HERE
            con.getInputStream().close();
            return readHttpInputStream(newUrl, headers, payload);
        }

        return new StreamConnection.UrlStreamConnection(con);
    }
```

common/src/test/java/apoc/util/UtilIT.java
gem-neo4j: setUpServer takes a redirect URL; in this case "http://127.168.0.1". Then getServerUrl() will call the Docker Containers IP (is assigned on startup. Then if the Dockers ip is called, the redirect command will happen, which returns a 301 followed by the input redirect URL, so it goes DOCKER_IP --> "http://127.168.0.1", the DOCKER_IP is not deny_listed so will pass, then the latter IP is deny_listed, so will fail

common/src/main/java/apoc/util/Util.java
```diff
@@ -119,6 +120,8 @@ public class Util {
     public static final String COMPILED = "interpreted"; // todo handle enterprise properly
     public static final String ERROR_BYTES_OR_STRING = "Only byte[] or url String allowed";
 
+    public static final int REDIRECT_LIMIT = 10;
```
nadja-muller, line 123: Sounds like something that we should document

nadja-muller (approved):

PR #244, opened 2022-11-18
[NwnPuJ91] Rewrite tests for failures in a safer way

Lojjs:
Tests with '(expected = QueryExecutionException.class)' or similar can
fail with the correct exception for unrelated reasons before coming to the assertion, making the test green even if it is broken. Also ignoreExceptions() feels dangerous to have in test setups 

Builds on top of https://github.com/neo4j/apoc/pull/240

### DIFF
--- a/core/src/test/java/apoc/atomic/AtomicTest.java
+++ b/core/src/test/java/apoc/atomic/AtomicTest.java
@@ -177,7 +177,7 @@ public class AtomicTest {
 		);
 		Throwable except = ExceptionUtils.getRootCause(e);
         assertTrue(except instanceof RuntimeException);
-		assertEquals("Attention your position out of range or higher than array length, that is 3", except.getMessage());
+		assertEquals("Position 5 is out of range for array of length 3", except.getMessage());
     }
 
     @Test
@@ -190,7 +190,7 @@ public class AtomicTest {
 		);
         Throwable except = ExceptionUtils.getRootCause(e);
         assertTrue(except instanceof RuntimeException);
-		assertEquals("Attention your position out of range or higher than array length, that is 0", except.getMessage());
+		assertEquals("Position 1 is out of range for array of length 0", except.getMessage());
     }
 
 	@Test
--- a/core/src/test/java/apoc/atomic/AtomicTest.java
+++ b/core/src/test/java/apoc/atomic/AtomicTest.java
@@ -2,11 +2,13 @@ package apoc.atomic;
 
 import apoc.util.TestUtil;
 import org.apache.commons.lang.ArrayUtils;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.hamcrest.Matchers;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.QueryExecutionException;
 import org.neo4j.graphdb.Relationship;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.test.rule.DbmsRule;
@@ -18,6 +20,8 @@ import java.util.concurrent.TimeUnit;
 
 import static apoc.util.MapUtil.map;
 import static apoc.util.TestUtil.testCall;
+import static junit.framework.TestCase.assertEquals;
+import static junit.framework.TestCase.assertTrue;
 import static org.junit.Assert.*;
 
 /**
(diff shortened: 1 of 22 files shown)


core/src/test/java/apoc/atomic/AtomicTest.java
```diff
@@ -161,19 +165,31 @@ public void testRemoveLastItemArray(){
...
-		assertThat(ArrayUtils.toObject(ages), Matchers.arrayContaining(40L, 50L, 60L));
+
+		QueryExecutionException e = assertThrows(QueryExecutionException.class,
+				() ->  testCall(db, "MATCH (n:Person {name:'Tom'}) CALL apoc.atomic.remove(n,$property,$position,$times) YIELD container RETURN count(*)",
+						map("property","age","position",5,"times",5), (r) -> {})
+
+		);
+		Throwable except = ExceptionUtils.getRootCause(e);
+        assertTrue(except instanceof RuntimeException);
+		assertEquals("Attention your position out of range or higher than array length, that is 3", except.getMessage());
```
gem-neo4j, line 179: Nothing to do with this PR, but maybe we need to fix the English of our APOC Error messages sometime 🙈

core/src/test/java/apoc/load/LoadJsonTest.java
```diff
@@ -321,19 +322,16 @@ public void testLoadJsonWithPathOptions() throws Exception {
...
-        }
+    @Test
+    public void testLoadJsonByUrlInConfigFileWrongKey() {
+        QueryExecutionException e = assertThrows(QueryExecutionException.class,
+                () -> testResult(db, "CALL apoc.load.json($key)",map("key","foo"), (r) -> r.hasNext())
+        );
+        Throwable except = ExceptionUtils.getRootCause(e);
+        assertTrue(except instanceof IOException);
+        final String message = except.getMessage();
+        assertTrue(message.startsWith("Cannot open file "));
```
gem-neo4j, line 333: Do we know what is in the middle of this string?

core/src/test/java/apoc/path/PathsTest.java
```diff
@@ -51,9 +51,14 @@ public void slice() throws Exception {
...
     }
 
-    @Test(expected = QueryExecutionException.class)
-    public void combineFail() throws Exception {
-        TestUtil.testCall(db, "MATCH p1 = (a:A)-[:NEXT]->(b:B), p2 = (c:C)-[:NEXT]->() RETURN apoc.path.combine(p1,p2) AS p", (row) -> fail("Should have failed"));
+    @Test
+    public void combineFail() {
+        QueryExecutionException e = assertThrows(QueryExecutionException.class,
+                () -> TestUtil.testCall(db, "MATCH p1 = (a:A)-[:NEXT]->(b:B), p2 = (c:C)-[:NEXT]->() RETURN apoc.path.combine(p1,p2) AS p",
+                        (row) -> fail("Should have failed"))
```
gem-neo4j, line 58: Do we need this row check?

core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -542,53 +547,62 @@ private long tryReadCount(int maxAttempts, String statement, long expected) thro
...
-    @Test(expected = QueryExecutionException.class, timeout = 1000)
+    @Test
     public void testIterateQueryFailInvalidConcurrency() {
         final String query = "CALL apoc.periodic.iterate('UNWIND range(0, 10) AS x RETURN x', " +
                 "'RETURN x', " +
                 "{concurrency:0 ,parallel:true})";
-        testFail(query);
+
+        QueryExecutionException e = assertThrows(QueryExecutionException.class,
+                () -> testCall(db, query, row -> fail("The test should fail but it didn't"))
```
gem-neo4j, line 590: Do we need this comment on the row?

core/src/test/java/apoc/util/UtilTest.java
```diff
@@ -79,14 +80,12 @@ public void mergeNullMaps() {
...
+    @Test
     public void testValidateQuery() {
-        try {
-            Util.validateQuery(db, "Match (n) return m");
-        } catch (QueryExecutionException e) {
-            assertTrue(e.getMessage().contains("Variable `m` not defined"));
-            throw e;
-        }
+        QueryExecutionException e = assertThrows(QueryExecutionException.class,
+                () -> Util.validateQuery(db, "Match (n) return m")
```
gem-neo4j, line 86: Not really a part of the PR, but it kinda bothers me our Cypher is written this way, would be nice to have it match the style guide, but super low prio I guess 🤷‍♀️

gem-neo4j (approved):

core/src/test/java/apoc/atomic/AtomicTest.java
Lojjs: I fixed this one as it was indeed very bad English

core/src/test/java/apoc/load/LoadJsonTest.java
Lojjs: It contains the full path which is user/CI dependent (in my case is  /Users/louise/Documents/neo4j-apoc-core/core/) so there is a good reason it doesn't check for the whole message

gem-neo4j: Ah that makes sense :) Thanks

core/src/test/java/apoc/util/UtilTest.java
Lojjs: Fixed this one, if you want to do it for more places you can create a DayOFCleanup card

PR #245, opened 2022-11-18
[iqpGvELg] Fix CWE-331: Insufficient Entropy in apoc random functions

vga91:
Changed `ThreadLocalRandom.current` with `java.security.SecureRandom`


nadja-muller:
From what I've read, we should consider using SecureRandomn only if we want to provide procedures that can be used in security-sensitive applications, such as cryptography, as this might have worse performance and ThreadLocalRandom is better when executing things in parallel. My guess would be that neither of these procedures are likely to be used for such a thing, but I've asked our security team to provide us with some guidelines whether we should or should not update this.

nadja-muller:
The verdict was that we should stick with ThreadLocalRandom for now. Thank you for looking into this regardless :)

PR #247, opened 2022-11-18
[Zl3h9sM8] apoc.schema.relationships return wrong output for relationship indexes

vga91:
For consistency with the `apoc. schema.nodes`, I put `"INDEX"` instead of relType.

I've put the old `"type"` in a new field `"relationshipType"`, in order not to lose output info and to be consistent with `"label"` in `schema.nodes` proc (I don't know if it can be considered an additional feature).

To discuss:
In any case, maybe, in both nodes and relationships procedure, instead of get only "INDEX" we could use `IndexDescriptor.getIndexType()` to get a more comprehensive "FULLTEXT"/"LOOKUP"/"TEXT"/"RANGE" or "POINT".
Most likely, since it's a feature, this change could be done in another card, in case.

### DIFF
--- a/common/src/main/java/apoc/result/IndexConstraintNodeInfo.java
+++ b/common/src/main/java/apoc/result/IndexConstraintNodeInfo.java
@@ -34,18 +34,18 @@ public class IndexConstraintNodeInfo {
      * @param label
      * @param properties
      * @param status status of the index, if it's a constraint it will be empty
-     * @param type if it is an index type will be "INDEX" otherwise it will be the type of constraint
+     * @param schemaType if it is an index type will be "INDEX" otherwise it will be the type of constraint
      * @param failure
      * @param populationProgress
      * @param size
      * @param userDescription
      */
-    public IndexConstraintNodeInfo(String name, Object label, List<String> properties, String status, String type, String failure, float populationProgress, long size, double valuesSelectivity, String userDescription) {
+    public IndexConstraintNodeInfo(String name, Object label, List<String> properties, String status, String schemaType, String failure, float populationProgress, long size, double valuesSelectivity, String userDescription) {
         this.name = name;
         this.label = label;
         this.properties = properties;
         this.status = status;
-        this.type = type;
+        this.type = schemaType;
         this.failure = failure;
         this.populationProgress = populationProgress;
         this.size = size;
--- a/common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
+++ b/common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
@@ -17,9 +17,9 @@ public class IndexConstraintRelationshipInfo {
     
     public final Object relationshipType;
 
-    public IndexConstraintRelationshipInfo(String name, String type, List<String> properties, String status, Object relationshipType) {
+    public IndexConstraintRelationshipInfo(String name, String schemaType, List<String> properties, String status, Object relationshipType) {
         this.name = name;
-        this.type = type;
+        this.type = schemaType;
         this.properties = properties;
         this.status = status;
         this.relationshipType = relationshipType;
--- a/common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
+++ b/common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
@@ -9,16 +9,19 @@ public class IndexConstraintRelationshipInfo {
 
     public final String name;
 
-    public final Object type;
+    public final String type;
 
     public final List<String> properties;
 
     public final String status;
+    
+    public final Object relationshipType;
 
-    public IndexConstraintRelationshipInfo(String name, Object type, List<String> properties, String status) {
+    public IndexConstraintRelationshipInfo(String name, String type, List<String> properties, String status, Object relationshipType) {
         this.name = name;
         this.type = type;
         this.properties = properties;
         this.status = status;
+        this.relationshipType = relationshipType;
     }
 }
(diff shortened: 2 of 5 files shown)


common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
```diff
@@ -9,16 +9,19 @@ public class IndexConstraintRelationshipInfo {
...
+    public final String type;
 
     public final List<String> properties;
 
     public final String status;
+    
+    public final Object relationshipType;
 
-    public IndexConstraintRelationshipInfo(String name, Object type, List<String> properties, String status) {
+    public IndexConstraintRelationshipInfo(String name, String type, List<String> properties, String status, Object relationshipType) {
```
gem-neo4j, line 20: can we call this `constraintType` then it is even less confusing, if you agree, change the one in Nodes as well :)

common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
```diff
@@ -9,16 +9,19 @@ public class IndexConstraintRelationshipInfo {
...
     public final String name;
 
-    public final Object type;
+    public final String type;
 
     public final List<String> properties;
 
     public final String status;
+    
+    public final Object relationshipType;
```
gem-neo4j, line 18: Why is this an Object? (I see it was before as well), but is it not just a string?

core/src/main/java/apoc/schema/Schemas.java
```diff
@@ -574,14 +574,15 @@ private IndexConstraintRelationshipInfo relationshipInfoFromIndexDescription(Ind
...
                 .collect(Collectors.toList());
         try {
-            return new IndexConstraintRelationshipInfo(getSchemaInfoName(relName, properties), relName, properties, schemaRead.indexGetState(indexDescriptor).toString());
+            return new IndexConstraintRelationshipInfo(getSchemaInfoName(relName, properties), "INDEX", properties, schemaRead.indexGetState(indexDescriptor).toString(), relName);
         } catch (IndexNotFoundKernelException e) {
             return new IndexConstraintRelationshipInfo(
                     // Pretty print for index name
                     getSchemaInfoName(relName, properties),
-                    relName,
+                    "INDEX",
```
gem-neo4j, line 582: above we check:  `!indexDescriptor.isUnique() ? "INDEX" : "UNIQUENESS"`, coming very soon is new relationship constraints that allow this, does this check work here too?

gem-neo4j (changes requested):
This looks nice overall, just a few comments! About the index type, I agree that this would be a new feature, and potentially even a breaking change, if you make a card for it, we can discuss it in refinement :)

core/src/main/java/apoc/schema/Schemas.java
vga91: Oh you're right, I didn't remember about your card, I was referring to the current version. Updated :)

common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
vga91: Because `relationshipType` can also be a `List<String>`, in case of composite indexes, e.g. in `testIndexesWithMultipleLabelsAndRelTypes` https://github.com/neo4j/apoc/pull/247/files/9b4b7bf4448cf27d6daff1f1bc9a9d44e23aa11e#diff-231d1cab850aaf45247b0fd9e1a64e18a8002685841df498e2098c38bedf6fb3R758

common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
vga91: Yes, changing the name is a good idea, but maybe I would change the name to `schemaType`, 
because `constraintType` = `INDEX` is personally confusing anyway.

It would seem clearer to me even if this change will be made: https://github.com/neo4j/apoc/pull/247#pullrequestreview-1188008107

common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
gem-neo4j: Ah makes sense, thanks :)

common/src/main/java/apoc/result/IndexConstraintRelationshipInfo.java
gem-neo4j: ah fair enough, schemaType works better :) just internally anyway

gem-neo4j (approved):

common/src/main/java/apoc/result/IndexConstraintNodeInfo.java
```diff
@@ -34,18 +34,18 @@ public class IndexConstraintNodeInfo {
      * @param label
      * @param properties
      * @param status status of the index, if it's a constraint it will be empty
-     * @param type if it is an index type will be "INDEX" otherwise it will be the type of constraint
+     * @param schemaType if it is an index type will be "INDEX" otherwise it will be the type of constraint
```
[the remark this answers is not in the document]
vga91: Actually not, because the output in handled by the field, so `public final String type;` in this case.
Now I have a doubt, is this what you meant by "internally"? 
If not, I can reset it :)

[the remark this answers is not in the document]
gem-neo4j: Oh, you're correct! This is good then :)

PR #250, opened 2022-11-22
[gRNAet7B] Fixes trigger procedures in clusters for neo4j 5.x

vga91:
- Cherry pick of https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3183 ,
- Fix flaky cluster test https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3323
- Fix other flakinesses https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3355
- Added trigger show procedure https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3335
- cherry pick of https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3369/files

-----

Related docs pr: https://github.com/neo4j/docs-apoc/pull/56


full/src/test/java/apoc/trigger/TriggerNewProceduresExtendedTest.java
```diff
@@ -0,0 +1,227 @@
...
+        db.executeTransactionally("MATCH (:Start)-[r:MY_TYPE]->(:End) DELETE r");
+
+        TestUtil.testCall(db, assertionQuery, (row) -> {
+            final Entity n = (Entity) row.get("n");
+            assertEquals("MY_TYPE", n.getProperty("type"));
+            assertEquals("val1", n.getProperty("prop1"));
+            assertEquals("val2", n.getProperty("prop2"));
+        });
+    }
+}
```
AzuObs, line 227: This is an Extended file and should not be cherry-picked to this repository

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -12,47 +13,42 @@
...
-
-        public TriggerInfo(String name, String query, Map<String, Object> selector, boolean installed, boolean paused) {
-            this.name = name;
-            this.query = query;
-            this.selector = selector;
-            this.installed = installed;
-            this.paused = paused;
-        }
+    // public for testing purpose
+    public static final String SYS_NON_WRITER_ERROR = "It's not possible to write into a cluster member with a non-writer role system database.\n";
```
AzuObs, line 24: I think this message could perhaps be improved because it doesn't really have anything to do with the "system" database AFAIK. 

"This instance is not allowed to write to the database because it is a follower. Please use these procedures against the database leader.\n"

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -121,15 +126,17 @@ public Stream<TriggerInfo> pause(@Name("name")String name) {
...
+        
         Map<String, Object> resume = triggerHandler.updatePaused(name, false);
 
         return Stream.of(new TriggerInfo(name,
                 (String)resume.get("statement"),
                 (Map<String,Object>) resume.get("selector"),
                 (Map<String,Object>) resume.get("params"),true, false));
     }
-
 }
```
AzuObs, line 142: In total the repository seems to have 3 methods called toTriggerInfo, two constructors for TriggerInfo, and also a method from which are all instantiaters for TriggerInfo. 

I haven't properly checked but this seems like quite a lot.
- Is there anyway this could be simplified? 
- Would it make sense to move them all to the TriggerInfo class to keep the code more maintainable?

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -18,41 +20,40 @@
...
-        public boolean paused;
-
-        public TriggerInfo(String name, String query, Map<String, Object> selector, boolean installed, boolean paused) {
-            this.name = name;
-            this.query = query;
-            this.selector = selector;
-            this.installed = installed;
-            this.paused = paused;
-        }
+    // public for testing purpose
```
AzuObs, line 23: This comment is probably not needed as it's fairly easy to find usages.

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -18,41 +20,40 @@
...
 
-    @Context public GraphDatabaseService db;
+    @Context public Log log;
 
-    @Context public TriggerHandler triggerHandler;
+    private void preprocessDeprecatedProcedures() {
+        final String msgDeprecation = """
+                Please note that the current procedure is deprecated,
+                it's recommended to use the `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, and `apoc.trigger.start` procedures,
+                instead of, respectively, `apoc.trigger.add`, `apoc.trigger.remove`, `apoc.trigger.removeAll`, `apoc.trigger.pause`, and `apoc.trigger.resume`.""";
```
AzuObs, line 40:
```suggestion
                Please note that the current procedure is deprecated.
                It is recommended to use the `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, and `apoc.trigger.start` procedures,
                instead of, respectively, `apoc.trigger.add`, `apoc.trigger.remove`, `apoc.trigger.removeAll`, `apoc.trigger.pause`, and `apoc.trigger.resume`.""";
```

AzuObs, line 40:
```suggestion
                Please note that the current procedure is deprecated.
                It is recommended to use the `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, and `apoc.trigger.start` procedures,
                instead of, respectively, `apoc.trigger.add`, `apoc.trigger.remove`, `apoc.trigger.removeAll`, `apoc.trigger.pause`, and `apoc.trigger.resume`.""";
```

AzuObs, line 40:
```suggestion
                Please note that the current procedure is deprecated.
                It is recommended to use the `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, and `apoc.trigger.start` procedures,
                instead of, respectively, `apoc.trigger.add`, `apoc.trigger.remove`, `apoc.trigger.removeAll`, `apoc.trigger.pause`, and `apoc.trigger.resume`.""";
```

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -18,41 +20,40 @@
...
-            this.query = query;
-            this.selector = selector;
-            this.installed = installed;
-            this.paused = paused;
-        }
+    // public for testing purpose
+    public static final String DB_NON_WRITER_ERROR = """
+            This instance is not allowed to write to the database because it is a follower.
+            Please use these procedures against the database leader.
+            """;
```
AzuObs, line 27:
```suggestion
    public static final String DB_NON_WRITER_ERROR = """
            This instance is not allowed to write to the database because it is a follower.
            Please use this procedure against the database leader.
            """;
```

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -18,41 +20,40 @@
...
-            this.paused = paused;
-        }
-    }
+    @Context public TriggerHandler triggerHandler;
 
-    @Context public GraphDatabaseService db;
+    @Context public Log log;
 
-    @Context public TriggerHandler triggerHandler;
+    private void preprocessDeprecatedProcedures() {
```
AzuObs, line 36: I think this function tries to do too much. I tries to log a warning saying the procedure is deprecated, and then it also does validates that we're allowed to write to the session database. 

I also think that the current logic is incomplete, because not only do we need to check that we're allowed to write to the session database, we also need to check that we're allowed to write to the system database.

I would recommend splitting organising the code around three functions:
- log warn message
- validate we're allowed to write to the session database
- validate we're allowed to write to the system database

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -18,41 +20,40 @@
...
+        }
+    }
+
+    @Admin
+    @Deprecated
+    @Procedure(name = "apoc.trigger.add", mode = Mode.WRITE, deprecatedBy = "apoc.trigger.install")
     @Description("Adds a trigger to the given Cypher statement.\n" +
             "The selector for this procedure is {phase:'before/after/rollback/afterAsync'}.")
-    public Stream<TriggerInfo> add(@Name("name") String name, @Name("statement") String statement, @Name(value = "selector"/*, defaultValue = "{}"*/)  Map<String,Object> selector, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
+    public Stream<TriggerInfo> add(@Name("name") String name, @Name("kernelTransaction") String statement, @Name(value = "selector"/*, defaultValue = "{}"*/)  Map<String,Object> selector, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
```
AzuObs, line 54:
```suggestion
    public Stream<TriggerInfo> add(@Name("name") String name, @Name("statement") String statement, @Name(value = "selector")  Map<String,Object> selector, @Name(value = "config", defaultValue = "{}") Map<String,Object> config) {
```

AzuObs: The word "statement" is used in the docs, and in general I think it's a slightly clearer variable name than "kernelTransaction".

core/src/main/java/apoc/trigger/Trigger.java
AzuObs: Generally speaking in formal wording you would want to use the short-forms:
- isn't
- it's
- doesn't
- mustn't

And instead you should wish to favour long-forms
- is not
- it is
- does not
- must not

Unfortunately I'm not sure how much we adhere to this when writing our documentation and error warnings 🤷

core/src/main/java/apoc/trigger/Trigger.java
AzuObs: I also wonder whether the warning message is needed. I've noticed that these procedures are annotated with 

```java
@Deprecated(name = foo, deprecatedBy = bar)
```

I wonder whether this is in itself logging a deprecation warning in the database response? Iff this is true, then this would be even better than logging the deprecation to the stdout file and at that point I would probably recommend not logging the warning to the stdout file since I don't think the dbms does that.

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME;
+
+
+public class TriggerNewProcedures {
+    // public for testing purpose
+    public static final String TRIGGER_NOT_ROUTED_ERROR = "The procedure should be routed and executed against a writer database";
```
AzuObs, line 25:
```suggestion
    public static final String TRIGGER_NOT_ROUTED_ERROR = "The procedure should be routed and executed against a writer database.";
```

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME;
+
+
+public class TriggerNewProcedures {
+    // public for testing purpose
```
AzuObs, line 24: Similar to another comment I made, where I don't think this comment is particularly useful because we can quite easily find usages.

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Context public GraphDatabaseAPI db;
+    
+    @Context public Log log;
+    
+    @Context public Transaction tx;
+
+    private void checkInSystemWriter() {
+        TriggerHandlerNewProcedures.checkEnabled();
+        // routing check
```
AzuObs, line 36: I'm not sure how useful this comment is, especially because the name of this `checkInSystemWriter` function is fairly descriptive about what it does

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,129 @@
...
+        if (!db.databaseName().equals(SYSTEM_DATABASE_NAME) || !Util.isWriteableInstance(db)) {
+            throw new RuntimeException(TRIGGER_NOT_ROUTED_ERROR);
+        }
+    }
+
+    private void checkTargetDatabase(String databaseName) {
+        if (databaseName.equals(SYSTEM_DATABASE_NAME)) {
+            throw new RuntimeException(TRIGGER_BAD_TARGET_ERROR);
+        }
```
AzuObs, line 45: I think this current check is a good start but I don't think it's sufficient. What we should is that we should also run a query "SHOW DATABASES" and check that the target database actually exists.

The user shouldn't be able to create a trigger for database "foo" if it doesn't exist. This logic was happening in the old procedures although it wasn't explicit. It was happening because in order to use the old triggers you had to connect to the session database and then validate the trigger statement - none of which was possible if the database didn't exist.

PS. You don't need to run the query "SHOW DATABASES" in a new system transaction because you're already in a system transaction when this code is executing.

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,129 @@
...
+        Map<String, Object> resume = TriggerHandlerNewProcedures.updatePaused(databaseName, name, false);
+
+        return Stream.of(TriggerInfo.from(resume, true));
+    }
+
+    // TODO - change with @SystemOnlyProcedure
+    @SystemProcedure
+    @Procedure(mode = Mode.READ)
+    @Description("CALL apoc.trigger.show(databaseName) | it lists all eventually installed triggers for a database")
+    public Stream<TriggerInfo> show(@Name("databaseName") String databaseName) {
```
AzuObs, line 121: It seems like we added the @Admin annotation to the apoc.trigger.list procedure in 4.4 but not the apoc.trigger.show procedure. I think that's a little inconsistent and should be changed in both 4.4 and 5.x so that all trigger procedures have the @Admin procedure since that seems to be the direction we decided to go.

core/src/main/java/apoc/trigger/TriggerHandlerNewProcedures.java
```diff
@@ -0,0 +1,160 @@
...
+                .filter(e -> !SystemPropertyKeys.database.name().equals(e.getKey()))
+                .collect(HashMap::new, // workaround for https://bugs.openjdk.java.net/browse/JDK-8148463
+                        (mapAccumulator, e) -> {
+                            Object value = List.of(SystemPropertyKeys.selector.name(), SystemPropertyKeys.params.name()).contains(e.getKey()) 
+                                    ? Util.fromJson((String) e.getValue(), Map.class) 
+                                    : e.getValue();
+                            
+                            mapAccumulator.put(e.getKey(), value);
+                    }, HashMap::putAll);
+    }
```
AzuObs, line 36: Could this go to TriggerInfo.java or would you prefer it stay in this file?

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,583 @@
...
+import static org.junit.Assert.*;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.neo4j.configuration.GraphDatabaseSettings.procedure_unrestricted;
+import static org.neo4j.internal.helpers.collection.MapUtil.map;
+import static org.neo4j.test.assertion.Assert.assertEventually;
+
+/**
+ * Test class for non-deprecated procedures, 
+ * i.e. `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, and `apoc.trigger.start`
+ */
```
AzuObs, line 47: Not sure how needed this comment is either, but if you'd like to keep this then could we perhaps also list the procedures apoc.trigger.show and apoc.trigger.list?

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,583 @@
...
+        awaitTriggerDiscovered(db, name, query);
+        
+        sysDb.executeTransactionally("CALL apoc.trigger.stop('neo4j', 'test')");
+        testCallEventually(db, "CALL apoc.trigger.list()", (row) -> {
+            assertEquals("test", row.get("name"));
+            assertEquals(true, row.get("installed"));
+            assertEquals(true, row.get("paused"));
+        }, TIMEOUT);
+    }
+
```
AzuObs, line 309: Could we add a test that checks the behaviour of the trigger is indeed pauses?

- Create Trigger "foo" which adds property "bar" to any node created
- Wait for "foo" to be installed
- Pause Trigger "foo"
- Wait for foo" to be paused
- Create node and make sure it doesn't have property "bar"

-

common/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -314,4 +316,4 @@ private void setLastUpdate(Transaction tx) {
         node.setProperty(SystemPropertyKeys.lastUpdated.name(), System.currentTimeMillis());
     }
 
-}
+}
```
AzuObs, line 319: I was wondering if you knew the reason why TriggerHandler is in `common` whereas all the other functionality (including TriggerHandlerNewProcedures) is in `core`?

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,583 @@
...
+                    Map<String, Object> row = res.next();
+                    assertEquals(name, row.get("name"));
+                    assertEquals(query, row.get("query"));
+                    row = res.next();
+                    assertEquals(name2, row.get("name"));
+                    assertEquals(query, row.get("query"));
+                    assertFalse(res.hasNext());
+                });
+    }
+
```
AzuObs, line 533: Could we add a test to make sure apoc.trigger.show only lists the triggers from the right database?

- create trigger for db "neo4j"
- create trigger for db "foo"
- show("neo4j") only shows neo4j
- show("foo") only shows foo

AzuObs: If you are happy to add this test, would you be happy to then back-port it to 4.4?

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,583 @@
...
+    
+    @Test
+    public void testInstallTriggerInUserDb() {
+        try {
+            testCall(db, "CALL apoc.trigger.install('neo4j', 'userDb', 'RETURN 1',{})",
+                    r -> fail("Should fail because of user db execution"));
+        } catch (QueryExecutionException e) {
+            assertTrue(e.getMessage().contains(TRIGGER_NOT_ROUTED_ERROR));
+        }
+    }
```
AzuObs, line 542: Could we add a test that installs a trigger in a user database which doesn't exist? The test should fail

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,583 @@
...
+
+    @Test
+    public void testInstallTriggerInSystemDb() {
+        try {
+            testCall(sysDb, "CALL apoc.trigger.install('system', 'name', 'SHOW DATABASES', {})", r -> fail(""));
+        } catch (RuntimeException e) {
+            assertTrue(e.getMessage().contains(TRIGGER_BAD_TARGET_ERROR));
+        }
+    }
+
```
AzuObs, line 552: Could we add a test where we start the dbms, then create a database "foo", then installs a trigger for "foo", and then check that the trigger actually works?

This is because we weren't sure whether whether the transaction event listeners would get registered for the database "foo" if it was created after the dbms starts up.

Hint: it should work fine, but I'd like a test to make it continues to work fine into the future

AzuObs: This is another one of those tests that might be good to back-port to 4.4 if you agree?

test-utils/src/main/java/apoc/trigger/TriggerTestUtil.java
```diff
@@ -0,0 +1,26 @@
...
+import org.neo4j.graphdb.GraphDatabaseService;
+
+import java.util.Map;
+
+import static apoc.util.TestUtil.testCallEventually;
+import static org.junit.Assert.assertEquals;
+
+public class TriggerTestUtil {
+    public static final long TIMEOUT = 10L;
+    public static final long TRIGGER_DEFAULT_REFRESH = 3000;
```
AzuObs, line 12: Could this be a bit lower perhaps? I don't see a reason to wait 3 seconds each time and it will only make the tests slower - I was thinking that perhaps 500 or 250 would be more appropriate.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @AfterClass
+    public static void bringDownCluster() {
+        if (cluster != null) {
+            cluster.close();
+        }
+    }
+    
+    // TODO: fabric tests
```
AzuObs, line 50:
```suggestion
    // TODO: fabric tests once the @SystemOnlyProcedure annotation is added to Neo4j
```

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -0,0 +1,178 @@
...
+                } catch (Exception e) {
+                    String errorMsg = e.getMessage();
+                    assertTrue("The actual message is: " + errorMsg, errorMsg.contains(triggerNotRoutedError));
+                }
+            }
+        }
+    }
+
+    @Test
+    public void testTriggersAllowedOnlyWithAdmin() {
```
AzuObs, line 107: Could this test be moved to TriggerTest.java and TriggerNewProceduresTest.java? Or else would you mind explaining why it needs to be run inside a cluster, as I'm not sure I know the reason?

AzuObs (commented):
My review has now been completed. 

Honestly this is a great cherry-pick! Unfortunately, I had a few nitpicks and also spotted a few issues that I might not have spotted in the 4.4 PR. I noticed that you also have a couple of failing trigger tests for some reason.

Keep up the good work, this will soon be over with 🙌

core/src/main/java/apoc/trigger/Trigger.java
vga91: Nothing seems to be logged with `@Deprecated`, the only difference I notice is this (on mouse over)

Schermata 2023-01-03 alle 12 25 20

But for example in cypher-shell, there's no way to check at a glance whether it's deprecated.

I added the system db check as well.

common/src/main/java/apoc/trigger/TriggerHandler.java
vga91: I honestly don't know, most likely an error when core and extended were split in 2 repos, probably because there is a `TriggerExtended`, which however does not use this class.

I moved this class (and also `TriggerMetadata`, placed in the same package),
and everything seems to be working fine.

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: The `TestDatabaseManagementServiceBuilder` is a community version so it doesn't support the "CREATE DATABASE dbName" command, nor the `DatabaseManagementService.createDatabase(..)` method.

Afaik I don't think it's easy (or feasible) to do it here. 
It would be better to create a new test class with an enterprise docker instance or reuse `TriggerClusterRoutingTest`.

And yes, we could back-port it to 4.4 as well.

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: Yes, it would be good to port it to 4.4. 

As above, I think it's better to use `TriggerClusterRoutingTest` or create another docker enterprise test

test-utils/src/main/java/apoc/trigger/TriggerTestUtil.java
vga91: I changed it [here](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3355/files#diff-346dafcb30d4bb1af05187a8d6c810d6bbf9779c222dd1c3c36bcfce44879d4aR74).

But, as mentioned there, I'm not sure why this thing seems to de-flake the test.

I could try to reduce the refresh here, maybe in 5.x it's not flaky.

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: Actually we don't need a cluster.
I placed this test here because this way we can do `GraphDatabase.driver(uri, AuthTokens.basic(nonAdminUser, nonAdminPwd)))`.

I don't think there is an easy way to do something like that in `TriggerNewProceduresTest` (but I might be wrong) which use a `TestDatabaseManagementServiceBuilder`.

We could also create another single-instance docker test (as mentioned above).

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
vga91: Added

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
vga91: My bad :( 
In 4.4 is present, for some reason I didn't put it here as well. Added

core/src/main/java/apoc/trigger/Trigger.java
AzuObs:
```json
{   "query": {     "text": "CALL apoc.warmup.run(false, false, false)",     "parameters": {}   },   "queryType": "r",   "counters": {     "_stats": {       "nodesCreated": 0,       "nodesDeleted": 0,       "relationshipsCreated": 0,       "relationshipsDeleted": 0,       "propertiesSet": 0,       "labelsAdded": 0,       "labelsRemoved": 0,       "indexesAdded": 0,       "indexesRemoved": 0,       "constraintsAdded": 0,       "constraintsRemoved": 0     },     "_systemUpdates": 0   },   "updateStatistics": {     "_stats": {       "nodesCreated": 0,       "nodesDeleted": 0,       "relationshipsCreated": 0,       "relationshipsDeleted": 0,       "propertiesSet": 0,       "labelsAdded": 0,       "labelsRemoved": 0,       "indexesAdded": 0,       "indexesRemoved": 0,       "constraintsAdded": 0,       "constraintsRemoved": 0     },     "_systemUpdates": 0   },   "plan": false,   "profile": false,   "notifications": [     {       "code": "Neo.ClientNotification.Statement.FeatureDeprecationWarning",       "title": "This feature is deprecated and will be removed in future versions.",       "description": "The query used a deprecated procedure. ('apoc.warmup.run' has been replaced by 'Firstly, the procedure duplicates functionality of page cache warm up which is a part of the DBMS. Secondly, the API of this procedure is very specific to Record storage engine.')",       "severity": "WARNING",       "position": {         "offset": 0,         "line": 1,         "column": 1       }     }   ],   "server": {     "address": "localhost:7687",     "agent": "Neo4j/5.1.0",     "protocolVersion": 5   },   "resultConsumedAfter": {     "low": 13,     "high": 0   },   "resultAvailableAfter": {     "low": 3,     "high": 0   },   "database": {     "name": "neo4j"   } }
```

The Neo4j DBMS has the ability to send back notifications in the query summary when returning query results to a user query. Whenever @Deprecated annotation is use, the Neo4j DBMS will send back a warning notification in the response for the overall query.

So I don't think we need to also be logging a warning message to the neo4j log file.

AzuObs: _PS. This is what the notification looks like in the browser (it also shows the squiggly lines which you noticed in your screenshot)._

Screenshot 2023-01-04 at 11 48 11

AzuObs: Btw, I made a mistake when I said we need to check that we need to do both:
- validate we're allowed to write to the session database
- validate we're allowed to write to the system database

I don't think we need to validate we're allowed to write to the session database, because both the new or the old procedures don't do that. Do they?

We only need to validate we're allowed to write to the system database, which is what the 4.4 behaviour does :+1:

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME;
+
+
+public class TriggerNewProcedures {
+    public static final String TRIGGER_NOT_ROUTED_ERROR = "The procedure should be routed and executed against a writer database.";
+    public static final String TRIGGER_BAD_TARGET_ERROR = "Triggers can only be installed on user databases.";
+    public static final String DB_NOT_FOUND_ERROR = "The user database with name '%s' does not exist";
```
AzuObs, line 28:
```suggestion
    public static final String DB_NOT_FOUND_ERROR = "The database with name '%s' does not exist";
```

common/src/main/java/apoc/trigger/TriggerHandler.java
AzuObs: You may need to check that APOC Extended still builds successfully after moving it. But I don't see why the functions in Extended would depend on this TriggerHandler file.

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -0,0 +1,596 @@
...
+import static org.junit.Assert.*;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.neo4j.configuration.GraphDatabaseSettings.procedure_unrestricted;
+import static org.neo4j.internal.helpers.collection.MapUtil.map;
+import static org.neo4j.test.assertion.Assert.assertEventually;
+
+/**
+ * Test class for non-deprecated procedures, 
+ * i.e. `apoc.trigger.install`, `apoc.trigger.drop`, `apoc.trigger.dropAll`, `apoc.trigger.stop`, `apoc.trigger.start`, `apoc.trigger.show` and `apoc.trigger.list`
+ */
```
AzuObs, line 48: The above comment is probably not needed

test-utils/src/main/java/apoc/trigger/TriggerTestUtil.java
AzuObs: Thanks for making the change. If we notice that it's flaky then we could potentially either investigate why a lower refresh rate is problematic, or we could bump this value back up!

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
AzuObs: There are 3 comments where we basically touch on this idea of "we should use the clustering test file, or we should make a new test file with an enterprise container".

I think for all 3 comments it would be nicer to not use the cluster, because it's incredibly slow to spin up and is unfortunately quite flaky. I think instead we should either modify the existing TriggerNewProceduresTest.java file to be dockerized neo4j rather than use the in-memory neo4j , or that we should make a new test file for enterprise tests which uses a dockerized neo4j.

core/src/main/java/apoc/trigger/Trigger.java
```diff
@@ -12,47 +14,53 @@
...
-
-        public TriggerInfo(String name, String query, Map<String, Object> selector, boolean installed, boolean paused) {
-            this.name = name;
-            this.query = query;
-            this.selector = selector;
-            this.installed = installed;
-            this.paused = paused;
+    private static final String NON_WRITER_ERROR = """
+            This instance is not allowed to write to the %1$s database.
+            Please use this procedure against a %1$s database writer.
```
AzuObs, line 27:
```suggestion
            Please open a session against a %1$s database writer when using this procedure.
```

AzuObs (commented):
Round 2 of reviewing is complete. Thank you for all the hard work and please let me know when you'd like me to take another look!

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: Deleted

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
vga91: Moved to `TriggerEnterpriseFeaturesTest`

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: Added to `TriggerEnterpriseFeaturesTest`

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
vga91: I added a new test `TriggerEnterpriseFeaturesTest`. 
Personally I preferred to leave the test as before for consistency with the `TriggerTest` (so we are sure that everything works correctly with neo4j community too).
But mostly because it's lighter. 

Also other tests, like `apoc.schema.`, have 2 test classes, 
one with community and the other with enterprise (only with features not available in community)

core/src/main/java/apoc/trigger/Trigger.java
vga91: Yes, that makes sense. Deleted both log and session check

AzuObs (approved):
Thank you so much for all the hard work. It looks great 💃

PR #254, opened 2022-11-25
[vULR7sSU] Update APOC descriptions

gem-neo4j:
Update missed and malformed descriptions

### DIFF
--- a/core/src/main/java/apoc/coll/Coll.java
+++ b/core/src/main/java/apoc/coll/Coll.java
@@ -362,7 +362,8 @@ public class Coll {
     }
 
     @UserFunction("apoc.coll.partition")
-    @Description("apoc.coll.partition(list,batchSize)")
+    @Description("Partitions the original list into sub-lists of the given batch size.\n" +
+            "The final list may be smaller than the given batch size.")
     public List<Object> partitionFn(@Name("values") List<Object> list, @Name("batchSize") long batchSize) {
         if (list==null || list.isEmpty()) return new ArrayList<>();
         return partitionList(list, (int) batchSize).collect(Collectors.toList());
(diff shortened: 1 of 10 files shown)


core/src/main/java/apoc/nodes/Nodes.java
```diff
@@ -653,20 +653,20 @@ public boolean isDense(@Name("node") Node node) {
     }
 
     @UserFunction("apoc.any.isDeleted")
-    @Description("returns boolean value for nodes and rele existance")
-    public boolean isDeleted(@Name("thing") Object thing) {
-        if (thing == null) return true;
+    @Description("Returns true if the given node or relationship no longer exists.")
```
gem-neo4j, line 656: This is just my copy for it, I'll get Jens to double check it (the original is so bad I couldn't leave it 🙈)

core/src/main/java/apoc/coll/Coll.java
```diff
@@ -963,13 +964,10 @@ public List<List<Object>> pairWithOffsetFn(@Name("coll") List<Object> values, @N
         return result;
     }
 
-    @Procedure
-    @Description("apoc.coll.pairWithOffset(values, offset) - returns a list of pairs defined by the offset")
+    @Procedure("apoc.coll.pairWithOffset")
+    @Description("Returns a list of pairs defined by the offset.")
     public Stream<ListResult> pairWithOffset(@Name("values") List<Object> values, @Name("offset") long offset) {
```
Lojjs, line 969: I noticed that the first parameter is named `coll` in the 3 different places it is listed in docs, but `values` here, and in the signature and input parameters on the docs page. Same applies to `apoc.coll.partition` above.

Not sure which way we would like to change it, but should be consistent.

core/src/main/java/apoc/nodes/Nodes.java
Lojjs: So this one was not documented at all?

Lojjs (commented):
Just one comment which might be more of a docs thing

core/src/main/java/apoc/load/LoadArrow.java
```diff
@@ -109,7 +109,7 @@ public Stream<MapResult> stream(
     }
 
     @Procedure(name = "apoc.load.arrow")
-    @Description("apoc.load.arrow(fileName, config) - imports nodes and relationships from the provided file")
+    @Description("Imports nodes and relationships from the provided arrow file.")
```
Lojjs, line 112: The docs PR did not have the word 'arrow' there if you want to be consistent, but on the other hand I think it is good to include it...

gem-neo4j: Ah hehe I'll make a PR to docs for this as I think it is good :)

Lojjs (approved):

PR #255, opened 2022-11-25
Code cleanup

gem-neo4j:
Cleanup some dead code and unused imports to reduce the crazy number of warnings.

There was also some deprecated usages of new Integer, new Long etc that I converted (first commit), these were considered errors and they are marked for removal.

This PR reduces our warnings by about 1000! (out of nearly 4000 though)


Lojjs (approved):
Great job 🧹

PR #256, opened 2022-11-28
[HFWBmuq2] Fixes #155: Check for correct tx terminations

### LINKED ISSUE #155: check for tx termination
<a href="https://github.com/jexp"></img></a> **Issue by [jexp](https://github.com/jexp)**
_Thursday Apr 06, 2017 at 11:37 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/375_

----

- [ ] Periodic
- [ ] Graph Refactorings
- [ ] The new renaming methods
- [ ] cypher.runFile
- [ ] exports
- [ ] load.json, load.jdbc, load.xml

In a sensible place in the outmost loop that's iterate a reasonable number of times.

vga91:
Fixes #155

- Added `terminationGuard.check()` where needed.
- Added various tests via `TransactionTestUtil` to verify procedures/functions are correctly terminated when the transaction is interrupted.
- Changed `movies.cypher` in order to be executed multiple times in `BigGraphTest`

### DIFF
--- a/common/src/main/resources/movies.cypher
+++ b/common/src/main/resources/movies.cypher
@@ -1,11 +1,11 @@
 CREATE (TheMatrix:Movie {title:'The Matrix', released:1999, tagline:'Welcome to the Real World'})
-CREATE (Keanu:Person {name:'Keanu Reeves', born:1964})
-CREATE (Carrie:Person {name:'Carrie-Anne Moss', born:1967})
-CREATE (Laurence:Person {name:'Laurence Fishburne', born:1961})
-CREATE (Hugo:Person {name:'Hugo Weaving', born:1960})
-CREATE (AndyW:Person {name:'Andy Wachowski', born:1967})
-CREATE (LanaW:Person {name:'Lana Wachowski', born:1965})
-CREATE (JoelS:Person {name:'Joel Silver', born:1952})
+CREATE (Keanu:Person:Other {name:'Keanu Reeves', id: randomUUID(), born:1964})
+CREATE (Carrie:Person:Other {name:'Carrie-Anne Moss', id: randomUUID(), born:1967})
+CREATE (Laurence:Person:Other {name:'Laurence Fishburne', id: randomUUID(), born:1961})
+CREATE (Hugo:Person:Other {name:'Hugo Weaving', id: randomUUID(), id: randomUUID(), born:1960})
+CREATE (AndyW:Person:Other {name:'Andy Wachowski', id: randomUUID(), born:1967})
+CREATE (LanaW:Person:Other {name:'Lana Wachowski', id: randomUUID(), born:1965})
+CREATE (JoelS:Person:Other {name:'Joel Silver', id: randomUUID(), born:1952})
 CREATE
   (Keanu)-[:ACTED_IN {roles:['Neo']}]->(TheMatrix),
   (Carrie)-[:ACTED_IN {roles:['Trinity']}]->(TheMatrix),
@@ -15,7 +15,7 @@ CREATE
   (LanaW)-[:DIRECTED]->(TheMatrix),
   (JoelS)-[:PRODUCED]->(TheMatrix)
   
-CREATE (Emil:Person {name:"Emil Eifrem", born:1978})
+CREATE (Emil:Person:Other {name:"Emil Eifrem", id: randomUUID(), born:1978})
 CREATE (Emil)-[:ACTED_IN {roles:["Emil"]}]->(TheMatrix)
 
 CREATE (TheMatrixReloaded:Movie {title:'The Matrix Reloaded', released:2003, tagline:'Free your mind'})
@@ -39,9 +39,9 @@ CREATE
   (JoelS)-[:PRODUCED]->(TheMatrixRevolutions)
   
 CREATE (TheDevilsAdvocate:Movie {title:"The Devil's Advocate", released:1997, tagline:'Evil has its winning ways'})
-CREATE (Charlize:Person {name:'Charlize Theron', born:1975})
-CREATE (Al:Person {name:'Al Pacino', born:1940})
-CREATE (Taylor:Person {name:'Taylor Hackford', born:1944})
+CREATE (Charlize:Person:Other {name:'Charlize Theron', id: randomUUID(), born:1975})
+CREATE (Al:Person:Other {name:'Al Pacino', id: randomUUID(), born:1940})
+CREATE (Taylor:Person:Other {name:'Taylor Hackford', id: randomUUID(), born:1944})
 CREATE
   (Keanu)-[:ACTED_IN {roles:['Kevin Lomax']}]->(TheDevilsAdvocate),
   (Charlize)-[:ACTED_IN {roles:['Mary Ann Lomax']}]->(TheDevilsAdvocate),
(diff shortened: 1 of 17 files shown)


common/src/main/resources/movies.cypher
```diff
@@ -1,11 +1,11 @@
 CREATE (TheMatrix:Movie {title:'The Matrix', released:1999, tagline:'Welcome to the Real World'})
-CREATE (Keanu:Person {name:'Keanu Reeves', born:1964})
-CREATE (Carrie:Person {name:'Carrie-Anne Moss', born:1967})
-CREATE (Laurence:Person {name:'Laurence Fishburne', born:1961})
-CREATE (Hugo:Person {name:'Hugo Weaving', born:1960})
-CREATE (AndyW:Person {name:'Andy Wachowski', born:1967})
-CREATE (LanaW:Person {name:'Lana Wachowski', born:1965})
-CREATE (JoelS:Person {name:'Joel Silver', born:1952})
+CREATE (Keanu:Person:Other {name:'Keanu Reeves', id: randomUUID(), born:1964})
```
gem-neo4j, line 2: This is the same file used for apoc.examples.movies? we should perhaps not change it if it is as it would be a breaking change I imagine, perhaps make a new file for testing purposes instead

core/src/main/java/apoc/cypher/Timeboxed.java
```diff
@@ -52,6 +57,12 @@ public Stream<MapResult> runTimeboxed(@Name("statement") String cypher, @Name("p
                 txAtomic.set(innerTx);
                 Result result = innerTx.execute(cypher, params == null ? Collections.EMPTY_MAP : params);
                 while (result.hasNext()) {
+                    if (Util.transactionIsTerminated(terminationGuard)) {
+                        txAtomic.get().close();
+                        offerToQueue(queue, POISON, timeout);
```
gem-neo4j, line 62: I don't really understand the Posion part here, if the tx is terminated would we not want to just return the same warning as below? I might be missing some context here, so feel free to explain it to me :)

core/src/main/java/apoc/load/Xml.java
```diff
@@ -190,6 +194,7 @@ private boolean proceedReader(XMLStreamReader reader) throws XMLStreamException
     }
 
     private void handleNode(Deque<Map<String, Object>> stack, Node node, boolean simpleMode) {
+        terminationGuard.check();
```
gem-neo4j, line 197: If this check fails does it throw an exception? do we need to do anything, or just let it happen :) perhaps a comment saying what it does could be useful

core/src/test/java/apoc/cypher/CypherTest.java
```diff
@@ -158,6 +161,36 @@ public void testWithTimeout() {
...
+    public void testRunTimeboxedWithTerminationInnerTransaction() {
+        final String innerLongQuery = "CALL apoc.util.sleep(10999) RETURN 0";
+        final String query = "CALL apoc.cypher.runTimeboxed($innerQuery, null, 99999)";
+
+        terminateTransactionAsync(db, 10L, innerLongQuery);
+        
+        // assert query terminated (RETURN 0)
+        TestUtil.testCall(db, query,
+                Map.of("innerQuery", innerLongQuery),
+                row -> assertEquals(Map.of("0", 0L), row.get("value")));
```
gem-neo4j, line 180: Is there a way to check warnings returned maybe? (if there is one returned that is)

core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -216,6 +219,42 @@ public void testTerminateIterate() throws Exception {
...
+
+        // assert query terminated (RETURN 0 nodesCreated)
+        try {
+            TestUtil.testCall(db, query,
+                    Map.of("innerQuery", innerLongQuery),
+                    row -> {
+                        final Object actual = ((Map) row.get("updateStatistics")).get("nodesCreated");
+                        assertEquals(0L, actual);
+                    });
+            fail("Should have terminated");
```
gem-neo4j, line 250: What does this fail give us?

test-utils/src/main/java/apoc/periodic/PeriodicTestUtils.java
```diff
@@ -41,8 +42,19 @@ public static boolean terminateQuery(String pattern, GraphDatabaseAPI db) {
         return numberOfKilledTransactions > 0;
     }
 
+    // todo - we could get rid of this method (and related code), and leverage the TransactionTestUtil.testTerminateWithCommand
```
gem-neo4j, line 45: I'd prefer todos are added as Trello tickets instead of in the code :)

test-utils/src/main/java/apoc/util/TransactionTestUtil.java
```diff
@@ -0,0 +1,95 @@
...
+                            "WHERE currentQuery CONTAINS $query AND NOT currentQuery STARTS WITH $transactionList " +
+                            "RETURN transactionId",
+                    map("query", query, "transactionList", TRANSACTION_LIST),
+                    result -> {
+                        final ResourceIterator<String> msgIterator = result.columnAs("transactionId");
+                        if (!msgIterator.hasNext()) {
+                            return false;
+                        }
+                        transactionId[0] = msgIterator.next();
+                        return transactionId[0] != null;
```
gem-neo4j, line 86: Why is the transactionId an array and not just a string?

core/src/test/java/apoc/load/LoadJsonTest.java
```diff
@@ -414,4 +421,10 @@ public void testLoadJsonParams() throws Exception {
                     assertFalse("value should be not empty", value.isEmpty());
                 });
     }
+
+    @Test
+    public void shouldTerminateLoadWhenTransactionIsTimedOut() {
+        final String query = "CALL apoc.load.json('https://github.com/knowitall/yelp-dataset-challenge/blob/master/data/yelp_phoenix_academic_dataset/yelp_academic_dataset_review.json?raw=truehttps://github.com/knowitall/yelp-dataset-challenge/blob/master/data/yelp_phoenix_academic_dataset/yelp_academic_dataset_review.json?raw=true')";
```
gem-neo4j, line 427: I'd prefer you didn't reference external resources in case it is deleted in the future, is it possible to add to this repo instead?

common/src/main/resources/movies.cypher
vga91: Oh, i didn't noticed it :( Now I created another file

core/src/main/java/apoc/cypher/Timeboxed.java
vga91: A bit strange but without the poison queue, the transaction is marked as terminated but actually continues the execution until it finishes, and after that throws an exception `The transaction has been terminated...`

Anyway, I realized that the `TransactionTestUtil.checkTerminationGuard` could potentially produce a false positive, therefore I also added the execution time check (which in fact fails without this `POISON`)

core/src/main/java/apoc/load/Xml.java
vga91: Yes it throw an exception if the check fails. 
Do you mean add a comment for each piece of code with `terminationGuard.check()`?

However I personally don't think if it's really necessary add a comment, because it's already explained in the [org.neo4j.procedure.TerminationGuard interface](https://github.com/neo4j/neo4j/blob/5.1/community/procedure-api/src/main/java/org/neo4j/procedure/TerminationGuard.java#L29)

core/src/test/java/apoc/cypher/CypherTest.java
vga91: In this case the `log.warn("query " + cypher + " has been terminated")` is not returned, because we terminate the inner transaction, so it doesn't execute that piece of code

core/src/test/java/apoc/periodic/PeriodicTest.java
vga91: It is just to verify that the query really fails, because of terminationGuard, and don't go ahead without errors.

test-utils/src/main/java/apoc/periodic/PeriodicTestUtils.java
vga91: I added the trello ticket, id `aWY2zoX5` in `Low priority` column, 
but I can always move it elsewhere, in case :)

core/src/main/java/apoc/load/Xml.java
gem-neo4j: Oh I didn't see the comment in the interface! This is fine then, thanks for explaining :)

test-utils/src/main/java/apoc/periodic/PeriodicTestUtils.java
gem-neo4j: Awesome, thanks

core/src/test/java/apoc/cypher/CypherTest.java
gem-neo4j: Ah I see, thanks :)

gem-neo4j (approved):
Nice work with this :)

core/src/test/java/apoc/cypher/CypherTest.java
jexp: and afaik there is no way for procedures to add a warning to the Cypher ExecutionContext

jexp:
@gem-neo4j are we good to merge this one?

gem-neo4j:
@jexp yep :) after a green build it will be

PR #257, opened 2022-11-28
[ADJvoENI] apoc.schema.properties.distinct(Count) can hang

vga91:
The test cases about 2nd scenario (ie fulltext indexes) is to be completed,
because it is to understand if the solution is good or not,
as it is in limbo between a bugfix and a feature.
That is, currently fulltext indexes are just not supported, hence the infinite hang.

To solve I created several `if(isFulltexIndexes) { fullTextImplementation } else { currentImplementation}`.

Alternatively we can do something like `if(isFulltexIndexes) { skipCurrentFulltextIndex }`, and document this.

But I think this is not a correct behavior, because as a user I would expect `apoc.schema.properties` to support any type of index, so maybe this is a bugfix and not a feature.

Wdyt?

### DIFF
--- a/core/src/main/java/apoc/index/SchemaIndex.java
+++ b/core/src/main/java/apoc/index/SchemaIndex.java
@@ -5,12 +5,16 @@ import apoc.util.QueueBasedSpliterator;
 import apoc.util.Util;
 import apoc.util.collection.Iterables;
 import apoc.util.collection.Iterators;
+import org.neo4j.common.EntityType;
 import org.neo4j.exceptions.KernelException;
 import org.neo4j.graphdb.Label;
+import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.graphdb.schema.IndexDefinition;
+import org.neo4j.graphdb.schema.IndexType;
 import org.neo4j.internal.kernel.api.*;
 import org.neo4j.internal.schema.IndexDescriptor;
+import org.neo4j.internal.schema.SchemaDescriptor;
 import org.neo4j.internal.schema.SchemaDescriptors;
 import org.neo4j.kernel.api.KernelTransaction;
 import org.neo4j.kernel.impl.api.KernelStatement;
@@ -21,10 +25,13 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.Procedure;
 import org.neo4j.procedure.TerminationGuard;
+import org.neo4j.procedure.builtin.FulltextProcedures;
 import org.neo4j.values.storable.Value;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
@@ -87,11 +94,24 @@ public class SchemaIndex {
                     CursorFactory cursors = ktx.cursors();
 
                     int[] propertyKeyIds = StreamSupport.stream(indexDefinition.getPropertyKeys().spliterator(), false)
-                            .mapToInt(name -> tokenRead.propertyKey(name))
+                            .mapToInt(tokenRead::propertyKey)
                             .toArray();
-                    String label = Iterables.single(indexDefinition.getLabels()).name();
-                    var schema = SchemaDescriptors.forLabel(tokenRead.nodeLabel(label), propertyKeyIds);
-                    IndexDescriptor indexDescriptor = Iterators.single(schemaRead.index(schema));
+                    
+                    SchemaDescriptor schema;
+                    if (isFullText(indexDefinition)) {
+                        int[] labelIds = Iterables.stream(indexDefinition.getLabels())
+                                .mapToInt(lbl -> tokenRead.nodeLabel(lbl.name()))
+                                .toArray();
+                        schema = SchemaDescriptors.fulltext(EntityType.NODE, labelIds, propertyKeyIds);
+                    } else {
+                        String label = Iterables.single(indexDefinition.getLabels()).name();
+                        schema = SchemaDescriptors.forLabel(tokenRead.nodeLabel(label), propertyKeyIds);
+                    }
+
+                    IndexDescriptor indexDescriptor = Iterators.singleOrNull(schemaRead.index(schema));
+                    if (indexDescriptor == null) {
+                        return null;
+                    }
                     scanIndex(queue, indexDefinition, key, read, cursors, indexDescriptor, ktx);
                 }
             }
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/index/SchemaIndex.java
```diff
@@ -104,35 +124,78 @@ private void scanIndex(BlockingQueue<PropertyValueCount> queue, IndexDefinition
...
+                if (isFullText(indexDefinition)) {
+                    long nodeReference = cursor.nodeReference();
+                    FulltextProcedures.NodeOutput nodeOutput = FulltextProcedures.NodeOutput.forExistingEntityOrNull(tx, nodeReference, cursor.score());
+                    final Node node = nodeOutput.node;
+                    indexDefinition.getPropertyKeys()
+                            .forEach(propKey -> {
+                                final Object property = node.getProperty(propKey, null);
+                                if (property != null) {
+                                    fullTextMap.merge(property, 1, Integer::sum);
+                                }
```
Lojjs, line 157: I think this is wrong (also see comment in the test). Here you are getting all property values in the index even if you are asking for a specific property in the APOC procedure. 

What makes it even harder is that if you have a node like `(:Label2 {prop1:`BB`})`, that would be in the fulltext index (which have OR semantics) and have the right property but should not be returned by `CALL apoc.schema.properties.distinct(Label1, prop1)` because the node doesn't have Label1.

So we need some sort of filtering on both the label and the property.

Lojjs (changes requested):
I think it would be ok to add support for fulltext index (we have e.g. added support for new types of constraints recently), but the way you have implemented it will not work. I would say in case you manage to write a good filtering step, we can add the support, otherwise we can document that it is not supported and skip it rather than hang the procs. 

Also had some ideas for tests and documentation, where the docs things can be added to the already existing Trello card

jexp:
@Lojjs @vga91 what's the status of this one?

Lojjs:
> @Lojjs @vga91 what's the status of this one?

I'm waiting for Giuseppe to respond to my review comments, but I think he had more pressing things to do around triggers

vga91:
Yes, I've started making the various changes, 
but at this moment I've left it aside to complete the pr of the triggers in 5.x.

I think I finish this pr in the day.

core/src/test/java/apoc/index/SchemaIndexTest.java
```diff
@@ -86,76 +102,191 @@ public void testDistinctPropertiesOnFirstIndex() {
...
                 });
 
-        db.executeTransactionally("DROP INDEX fulltextLabel1");
+        testCall(db, "CALL apoc.schema.properties.distinct($label, $key)",
+                map("label", SCHEMA_LABEL, "key", "prop2"),
+                row -> assertEquals(Set.of("bar", "foo"), Set.copyOf((List)row.get("value")))
+        );
+
+        testResult(db, SCHEMA_DISTINCT_COUNT_ORDERED,
+                map("label","","key",""),
```
Lojjs, line 165: So if the label or property is an empty string, that means all labels/properties? I think it would make sense to add that to the documentation as well, can add that to the existing documentation card

core/src/main/java/apoc/index/SchemaIndex.java
```diff
@@ -137,48 +144,40 @@ private void scanIndex(BlockingQueue<PropertyValueCount> queue, IndexDefinition
...
-                            previousValue = v;
-                            count = 1;
-                        }
+                final Node node = threadTx.getNodeById( cursor.nodeReference() );
+
+                labels.forEach(label -> {
+
+                    final boolean hasLabel = node.hasLabel(label);
+                    
+                    final Object property = node.getProperty(key, null);
```
Lojjs, line 166: Don't know if it make a noticable difference, but I guess for a slightly better performance, you could move this line before the labels.foreach loop.

core/src/test/java/apoc/index/SchemaIndexTest.java
```diff
@@ -86,76 +102,191 @@ public void testDistinctPropertiesOnFirstIndex() {
...
-                map("label", "Label1", "key", "prop1"),
-                row -> assertEquals(Set.of("AA", "EE", "UU", "Ryan", "Michael", "abcde", "qwerty"), Set.copyOf((List)row.get("value")))
+                map("label", FULL_TEXT_LABEL, "key", "prop1"),
+                row -> assertEquals(Set.of("AA", "EE", "UU", "Ryan", "Michael"), Set.copyOf((List)row.get("value")))
         );
 
-        testResult(db, "CALL apoc.schema.properties.distinctCount($label, $key) " +
-                        "YIELD label, key, value, count RETURN * ORDER BY label",
-                map("label", "Label1", "key", "prop1"),
+        // in this case the procedure returns equal pairs of rows because we have 2 different indexes
```
Lojjs, line 214: This might be surprising for a user. Do you think we should try to filter out equal lines or should we just document it?

Lojjs (commented):
I think the behaviour is OK now, but there is still some edge cases users might be surprised over. Take a look at my new comments and let me know what you think. I can collect all docs improvements into a checklist on the already existing docs card if you like

core/src/main/java/apoc/index/SchemaIndex.java
```diff
@@ -104,35 +124,78 @@ private void scanIndex(BlockingQueue<PropertyValueCount> queue, IndexDefinition
...
             // however the index guarantees to be scanned in order unless
             // there are writes done in the same tx beforehand - which we don't do.
-            IndexReadSession indexSession = read.indexReadSession( indexDescriptor );
-            read.nodeIndexScan(indexSession, cursor, IndexQueryConstraints.unorderedValues());
+            
+            final IndexReadSession indexSession;
+            try {
+                indexSession = read.indexReadSession( indexDescriptor );
+            }catch (Exception e) {
+                // we skip indexScan if it's still populating
```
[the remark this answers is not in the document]
vga91: Sorry, I missed this point. I added it as a checklist's item

core/src/test/java/apoc/index/SchemaIndexTest.java
vga91: Yes, maybe it makes more sense to remove the duplicates to avoid any doubts. Updated

core/src/main/java/apoc/index/SchemaIndex.java
vga91: Changed

core/src/test/java/apoc/index/SchemaIndexTest.java
vga91: Sounds good, I added it [here](https://trello.com/c/MBiu8YIz/1827-improve-documentation-about-apocschemapropertiesdistinctcount) as a checklist.

Lojjs (approved):
I think this looks good now

PR #261, opened 2022-12-05
[wokdlhX3] Extract code from APOC core which is only used in APOC extended

Lojjs:
Should go in together with https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3361


test-utils/src/main/java/apoc/util/GoogleCloudStorageContainerExtension.java
```diff
@@ -11,6 +11,7 @@
 
 import static java.net.HttpURLConnection.HTTP_OK;
 
+@SuppressWarnings("unused") // used from extended
```
conker84, line 14: I would prefer moving this on extended as it's used only in there as you did for many other APIs

test-utils/src/main/java/apoc/util/MySQLContainerExtension.java
```diff
@@ -5,6 +5,7 @@
 
 import java.time.Duration;
 
+@SuppressWarnings("unused") // used from extended
```
conker84, line 8: Same as `GoogleCloudStorageContainerExtension`

common/src/main/java/apoc/result/VirtualPath.java
```diff
@@ -139,23 +139,4 @@ private void requireConnected(Relationship relationship) {
             throw new IllegalArgumentException("Relationship is not part of current path.");
         }
     }
-    
-    public static final class Builder {
```
conker84, line 143: I prefer to have it in there, now it's used only in `DataVirtualizationCatalog` but by design would be better having it in there; as in the future could be used in other use-cases

conker84 (changes requested):
Thank you so much @Lojjs, just a couple of things that I want to discuss with you

conker84:
@Lojjs can you please rebase it?

conker84 (approved):

conker84:
@Lojjs LGTM thank you so much!

PR #262, opened 2022-12-06
[Q5C5ZMtb] Try to fix flaky ExportS3PerformanceTest

vga91:
Reduced the dataset size.

The test sometimes fails with an error:
```
apoc.export.ExportS3PerformanceTest > testExportAllCsvS3 STANDARD_OUT
    Data creation started.
Error: The operation was canceled.
```

Since `Data creation started.` is present, but not the next`Data creation finished.`,
I guess the error may be due to a heap space or timeout error.


AzuObs (approved):
I would be fine with trying this for now, but I have some doubts that this will work. It might make it less likely, but if the test keeps being flaky then we may need to dig deeper.

PR #265, opened 2022-12-08
[3KXXQazJ] Fix bug with serializing arrays of primitive values for json export query

gem-neo4j:
Fixes issue where a class cast exception stopped arrays of primitive values from serializing to JSON.

Also did some general code cleanup as I went 🧹


AzuObs (approved):
Great stuff 💯

PR #270, opened 2023-01-02
[FmCGhDLD] Remove relationship constraint feature flag

gem-neo4j:
The relationship uniqueness constraints are now turned on by default in Neo dev (feature flag set to true)

### DIFF
--- a/core/src/test/java/apoc/schema/SchemasTest.java
+++ b/core/src/test/java/apoc/schema/SchemasTest.java
@@ -16,7 +16,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import org.neo4j.configuration.GraphDatabaseInternalSettings;
 import org.neo4j.configuration.GraphDatabaseSettings;
 import org.neo4j.graphdb.Label;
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -49,8 +48,7 @@ public class SchemasTest {
 
     @Rule
     public DbmsRule db = new ImpermanentDbmsRule()
-            .withSetting(GraphDatabaseSettings.procedure_unrestricted, Collections.singletonList("apoc.*"))
-            .withSetting(GraphDatabaseInternalSettings.rel_unique_constraints, true);
```
nadja-muller, line 53: This is used in another class MetaTest.java. Should it be removed there, too?

nadja-muller (commented):
Looks good. I assume you want to remove the other occurence, too, though.

core/src/test/java/apoc/schema/SchemasTest.java
gem-neo4j: :O How did my search not find that 🙈 Thank you!

nadja-muller (approved):

PR #272, opened 2023-01-02
[RhUZ6yD8] Renaming for allowlist and denylist

gem-neo4j:
Add config values for allowlist and denylist and tests and deprecate the use of Whitelist and Blacklist.

### DIFF
--- a/core/src/test/java/apoc/path/ExpandPathTest.java
+++ b/core/src/test/java/apoc/path/ExpandPathTest.java
@@ -43,7 +43,7 @@ public class ExpandPathTest {
 
     @After
     public void removeOtherLabels() {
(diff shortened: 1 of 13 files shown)


core/src/test/java/apoc/path/ExpandPathTest.java
```diff
@@ -348,6 +348,136 @@ public void testCompoundLabelWorksInBlacklist() {
 				});
 	}
 
+	@Test
+	public void testSubgraphNodesWorksWithAllowList() {
+		TestUtil.testResult(db,
+				"""
+						MATCH (k:Person {name:'Keanu Reeves'})
+						MATCH (allowlist:Movie)
```
ncordon, line 356: Why do we need to test that we can use allowlist as a variable?

ncordon (approved):

PR #274, opened 2023-01-03
[MiVboeTJ] Fixes slf4j apoc logger

ncordon:
## Why
Because we were seeing messages like this one on startup:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

### DIFF
--- a/test-utils/build.gradle
+++ b/test-utils/build.gradle
@@ -18,15 +18,21 @@ dependencies {
         exclude group: 'org.apache.hive', module: 'hive-service'
     }
 
+    // The logger will come from the :common project via slf4j-simple
+    def withoutLoggers = {
+        exclude group: 'org.slf4j', module: 'slf4j-log4j12'
+        exclude group: 'org.slf4j', module: 'slf4j-reload4j'
+    }
+
     api group: 'junit', name: 'junit', version: '4.13.2'
     api group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     api group: 'org.hamcrest', name: 'hamcrest-library', version: '2.2'
     api group: 'org.neo4j.community', name: 'it-test-support', version: neo4jVersionEffective // , classifier: "tests"
     api group: 'org.neo4j', name: 'log-test-utils', version: neo4jVersionEffective // , classifier: "tests"
     api group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '4.0.0'
-    api group: 'org.apache.hadoop', name: 'hadoop-hdfs', version: '3.3.4', withoutServers
-    api group: 'org.apache.hadoop', name: 'hadoop-common', version: '3.3.4', withoutServers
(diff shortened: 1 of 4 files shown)


test-utils/build.gradle
```diff
@@ -18,15 +18,21 @@ dependencies {
...
     api group: 'junit', name: 'junit', version: '4.13.2'
     api group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     api group: 'org.hamcrest', name: 'hamcrest-library', version: '2.2'
     api group: 'org.neo4j.community', name: 'it-test-support', version: neo4jVersionEffective // , classifier: "tests"
     api group: 'org.neo4j', name: 'log-test-utils', version: neo4jVersionEffective // , classifier: "tests"
     api group: 'org.neo4j.driver', name: 'neo4j-java-driver', version: '4.0.0'
-    api group: 'org.apache.hadoop', name: 'hadoop-hdfs', version: '3.3.4', withoutServers
-    api group: 'org.apache.hadoop', name: 'hadoop-common', version: '3.3.4', withoutServers
-    api group: 'org.apache.hadoop', name: 'hadoop-minicluster', version: '3.3.4', withoutServers
+    api group: 'org.apache.hadoop', name: 'hadoop-hdfs', version: '3.3.4', withoutServers.andThen(withoutLoggers)
```
nadja-muller, line 33: Where does the other version come from?

nadja-muller (approved):

michael-simons:
Looking at the test code that has been added as well, a probably less invasive choice would have been adding the simple implementation to test scope only.

Now, APOC enforces that dependency to Neo4j.
Neo4j embedded enforces it to whatever container it is run in.
Somebodies application is using a different provider all of sudden.

See for reference
https://www.slf4j.org/codes.html#multiple_bindings

> Embedded components such as libraries or frameworks should not declare a dependency on any SLF4J binding but only depend on slf4j-api. When a library declares a compile-time dependency on a SLF4J binding, it imposes that binding on the end-user, thus negating SLF4J's purpose. When you come across an embedded component declaring a compile-time dependency on any SLF4J binding, please take the time to contact the authors of said component/library and kindly ask them to mend their ways.

In essence, it's kind of a hard problem, as usually Neo4j is the host for plugins, but can itself be hosted too. 
Probably it's worth doing a coordinated approach to solve this

- Neo4j as a standalone product should (and most likely must) contain a preferred implementation of SLF4j, so that the product has control over its logging flow
- Neo4j as an embeddable (both as embedded database but also as test-harness) should not do this, so that the above recommendation from SLF4J is upheld. Completely ok to default to no-op here.
- APOC and anything else that are only meant to be hosted or embedded must not carry implementations with them.

PR #275, opened 2023-01-04
[8JV7MoEx] Remove EOL versions from APOC readme and fix broken image links

gem-neo4j:
The versions below 4.4 are no longer supported, and it clutters up the readme a bit. Also the dev branch (default) never shows the images in the readme due to how it is set up, so just updated that as well to look better :)


AzuObs (approved):
Great stuff. Thanks for the cleanup job! 🎉

PR #285, opened 2023-01-13
[480BDuYi] Try de-flake some trigger tests with embedded db

vga91:
In 4.4 [I increased this value](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3355), and in this way the problem seems to be resolved.

In 5.x [I tried to put a lower value](https://github.com/neo4j/apoc/pull/250#discussion_r1060649214), 
but apparently some tests are flaky.

As with 4.4, I couldn't figure out exactly why this stabilize the tests (hoping it's for the same reason as 4.4), 
also because the problem seems to occur only with the embedded instance 
(i.e. `org.neo4j.test.TestDatabaseManagementServiceBuilder` )


AzuObs (approved):
Oki this is what we talked about :) Let's try this! Thanks for de-flaking the tests.

PR #287, opened 2023-01-16
[126ZorO2] Document that apoc.util.sha1 and apoc.text.random are not suitable for cryptographic use.

gem-neo4j:
During the Threat Modelling it was decided that apoc.util.sha1 and apoc.text.random should be documented as unsafe for cryptographic use to discourage users from thinking they are secure.


AzuObs (dismissed):
LGTM :)

AzuObs (approved):
LGTM :)

PR #288, opened 2023-01-16
[gcL5e77c] Fix edge cases in trigger methods

vga91:
- The `apoc.trogger.drop`, `apoc.trogger.stop` and `apoc.trogger.start` return an empty result if the trigger doesn't exist
- The `apoc.trigger.install` returns the current trigger instead of the previous one


AzuObs (approved):
LGTM :+1:

PR #292, opened 2023-01-18
[Ci2sJ23z] apoc.trigger.show doesn't work in a system follower

vga91:
- Removed check `isWriteableInstance` for `apoc.trigger.show` procedure.

- Removed unused `log`context

---

The `NON_SYS_DB_ERROR` message error may be changed, depending on the card `[mcub3FQf]`


core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -23,20 +22,23 @@
...
         
-        if (!db.databaseName().equals(SYSTEM_DATABASE_NAME) || !Util.isWriteableInstance(db)) {
+        if (!db.databaseName().equals(SYSTEM_DATABASE_NAME)) {
+            throw new RuntimeException(NON_SYS_DB_ERROR);
+        }
+        
+        if (checkWriteable && !Util.isWriteableInstance(db)) {
             throw new RuntimeException(TRIGGER_NOT_ROUTED_ERROR);
         }
     }
```
ncordon, line 44: Don't you think it would be better if we just changed the logic of the show method? This changes the logic for the rest of the trigger methods as well, so they would output a different error message now if they are executed against the neo4j database. So in show, call `checkInSystem()` that checks we are in a system database. In the rest of the trigger databases, call `checkInSystemWriter`, that check we are in a system LEADER (i.e. leave it as it was)

ncordon (changes requested):
Please see comment

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -4,7 +4,6 @@
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.kernel.api.procedure.SystemProcedure;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
-import org.neo4j.logging.Log;
```
ncordon, line 7: 👍

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
vga91: @ncordon Yes that makes sense. Changed

ncordon (approved):

PR #303, opened 2023-01-25
[ROQ0fkdz] Reduce the number of transactions we use in triggers

vga91:
The ideal solution I suppose is to use `@Context public Transaction tx`, which calls `ProcedureTransactionImpl`.

But with both this and `@Context publicGraphDatabaseAPI` [which calls ProcedureGraphDatabaseAPI] via `db.beginTx()`,

I receive the following error if I login with an user with role `admin` 
(both with the embedded db and with an enterprise instance):
```
Creating new node label on database 'system' is not allowed for user '' 
with FULL overridden by READ overridden by READ. See GRANT CREATE NEW NODE LABEL ON DATABASE system
```

Btw, even trying to explicit `GRANT CREATE NEW NODE LABEL ON DATABASE system TO admin` gives the same result.

I also tried changing `mode = Mode.WRITE` to `Mode.DBMS` but it doesn't change anything.

I think the above one might be a bug, or so, not APOC related.

---

The only working way is to use `@Context public GraphDatabaseAPI`, which calls `GraphDatabaseFacade`.

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerNewProcedures.java
+++ b/core/src/main/java/apoc/trigger/TriggerNewProcedures.java
@@ -33,10 +33,10 @@ public class TriggerNewProcedures {
     
     @Context public Transaction tx;
 
-    private void checkInSystemWriter(Transaction tx) {
+    private void checkInSystemWriter() {
         TriggerHandlerNewProcedures.checkEnabled();
         
-        if (!db.databaseName().equals(SYSTEM_DATABASE_NAME) || !Util.isWriteableInstance(db, tx)) {
+        if (!db.databaseName().equals(SYSTEM_DATABASE_NAME) || !Util.isWriteableInstance(db)) {
             throw new RuntimeException(TRIGGER_NOT_ROUTED_ERROR);
         }
     }
@@ -49,7 +49,7 @@ public class TriggerNewProcedures {
         }
     }
 
-    private void checkTargetDatabase(String databaseName, Transaction tx) {
+    private void checkTargetDatabase(String databaseName) {
         final Set<String> databases = tx.execute("SHOW DATABASES", Collections.emptyMap())
                 .<String>columnAs("name")
                 .stream()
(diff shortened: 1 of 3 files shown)


core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -48,7 +49,7 @@ private void checkInSystem() {
         }
     }
 
-    private void checkTargetDatabase(String databaseName) {
+    private void checkTargetDatabase(String databaseName, Transaction tx) {
```
ncordon, line 52: You don't need the transaction here I think? Because it's the same one from the class

core/src/main/java/apoc/trigger/TriggerNewProcedures.java
```diff
@@ -68,12 +69,15 @@ private void checkTargetDatabase(String databaseName) {
...
-        checkInSystemWriter();
-        checkTargetDatabase(databaseName);
-        Map<String,Object> params = (Map)config.getOrDefault("params", Collections.emptyMap());
-        
-        TriggerInfo triggerInfo = TriggerHandlerNewProcedures.install(databaseName, name, statement, selector, params);
-        return Stream.of(triggerInfo);
+
+        return withTransaction(tx -> {
+            checkInSystemWriter(tx);
+            checkTargetDatabase(databaseName, tx);
```
ncordon, line 75: For these two you don't need to create a new transaction, you can just recycle the one you are in, so I'd suggest reshaping it like:

```java
        checkInSystemWriter();
        checkTargetDatabase(databaseName);
        Map<String,Object> params = (Map)config.getOrDefault("params", Collections.emptyMap());

        return withTransaction(tx -> {
            TriggerInfo triggerInfo = TriggerHandlerNewProcedures.install(databaseName, name, statement, selector, params, tx);
            return Stream.of(triggerInfo);
        });
```

so you only open a new system transaction at the moment where you need it

vga91: Changed 👍

ncordon (approved):

PR #305, opened 2023-01-25
[wvzcpHQH] Add SSFR check for apoc.spatial procedures

vga91:
- Added failing tests for ip blocklist setting.

It's not necessary to change the implementation, 
as both `apoc.spatial.geocode` and `apoc.spatial.reverseGeocode` call `JsonUtil.loadJson(...)` 
which ends up on the work done [in this pr](https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/2649).

### DIFF
--- a/core/src/test/java/apoc/spatial/GeocodeTest.java
+++ b/core/src/test/java/apoc/spatial/GeocodeTest.java
@@ -2,7 +2,10 @@ package apoc.spatial;
 
 import apoc.util.JsonUtil;
 import apoc.util.TestUtil;
+import inet.ipaddr.IPAddressString;
 import org.junit.*;
+import org.neo4j.configuration.GraphDatabaseInternalSettings;
+import org.neo4j.graphdb.QueryExecutionException;
 import org.neo4j.test.rule.DbmsRule;
 import org.neo4j.test.rule.ImpermanentDbmsRule;
 
@@ -10,6 +13,7 @@ import java.io.InputStream;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import static apoc.ApocConfig.apocConfig;
 import static apoc.util.MapUtil.map;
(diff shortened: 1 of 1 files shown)


AzuObs (changes requested):
These tests do not seem to be passing in the CI or on my machine either. Would it be possible to fix them before resubmitting this for review?

vga91:
@AzuObs ok, now it seems to work in the CI as well

core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -34,8 +45,95 @@ public void testWrongUrlButViaOtherProvider() throws Exception {
         testGeocodeWithThrottling("osm", false, 
                 map("url", "https://api.opencagedata.com/geocode/v1/json?q=PLACE&key=KEY111"));
     }
+
+    @Test
+    public void testGeocodeWithBlockedAddressWithApocConf() {
+        final String geocodeBaseConfig = Geocode.PREFIX + "." + "opencage";
```
AzuObs, line 51:
```suggestion
        final String geocodeBaseConfig = Geocode.PREFIX + ".opencage";
```

core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -34,8 +45,95 @@ public void testWrongUrlButViaOtherProvider() throws Exception {
...
+
+            // check that if neither url nor reverse url are blocked 
+            // the procedures continue the execution (in this case by throwing a `java.net` tException)
+            apocConfig().setProperty(geocodeConfigUrl, nonBlockedUrl);
+            apocConfig().setProperty(geocodeConfigReverseUrl, nonBlockedReverseUrl);
+            
+            assertGeocodeFails(false, JAVA_NET_EXCEPTION);
+            assertGeocodeFails(true, JAVA_NET_EXCEPTION);
+        });
+    }
```
AzuObs, line 81: This test is taking about 1 minute to run. As if the test below. This seems a little bit excessive for these tests, do you think there is room to bring their execution down to ~few seconds?

core/src/test/java/apoc/spatial/GeocodeTest.java
```diff
@@ -34,8 +45,95 @@ public void testWrongUrlButViaOtherProvider() throws Exception {
...
+            
+            assertGeocodeFails(false, BLOCKED_ERROR);
+
+            // check that if neither url nor reverse url are blocked 
+            // the procedures continue the execution (in this case by throwing a `java.net` tException)
+            apocConfig().setProperty(geocodeConfigUrl, nonBlockedUrl);
+            apocConfig().setProperty(geocodeConfigReverseUrl, nonBlockedReverseUrl);
+            
+            assertGeocodeFails(false, JAVA_NET_EXCEPTION);
+            assertGeocodeFails(true, JAVA_NET_EXCEPTION);
```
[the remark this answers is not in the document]
AzuObs: Oh ok, I see what's happening. It fails but because of a different problem, which means that IP blocking worked. Perhaps this is a little awkward to understand, do you think there is a way we could make the query succeed when the IP is not blocked?

AzuObs (changes requested):
Thank you very much for fixing the tests. It's very pleasing to see that these tests clearly demonstrate that these procedures were already protected 🎉 

I left a couple of nitpicks regarding the time these tests take to run, and also regarding whether these tests could be more readable.

core/src/test/java/apoc/spatial/GeocodeTest.java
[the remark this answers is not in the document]
vga91: Unfortunately I don't think so, it would take an opencage api key to make that query succeed.
However I changed the name of these methods from `assertGeocodeFails ` to `assertGeocodeAllowedUrl`, at least to be a bit clearer.

core/src/test/java/apoc/spatial/GeocodeTest.java
vga91: I speeded up the queries a lot by changing the `NON_BLOCKED_ADDRESS` value from `123.456.7.8` to `localhost`. This way you still get a java.net exception but faster (because it's not a socketTimeoutException but a Connection refused)

AzuObs (approved):
Thank you for actioning the comments. Nice work 🙌

PR #306, opened 2023-01-25
[04b6IYYB] improve performance of apoc.meta procedures

nadja-muller:
APOC docs [here](https://github.com/neo4j/docs-apoc/pull/104). 

Make sampling start working as intended. Clean up a little bit how it all works and add tests.

### DIFF
--- a/core/src/main/java/apoc/meta/Meta.java
+++ b/core/src/main/java/apoc/meta/Meta.java
@@ -931,8 +931,8 @@ public class Meta {
             long sample = getSampleForLabelCount(labelCount, metaConfig.getSample());
             while (nodes.hasNext()) {
                 count++;
+                Node node = nodes.next();
                 if(count % sample == 0) {
-                    Node node = nodes.next();
                     long maxRels = metaConfig.getMaxRels();
                     for (Relationship rel : node.getRelationships(direction, relationshipType)) {
                         Node otherNode = direction == Direction.OUTGOING ? rel.getEndNode() : rel.getStartNode();
--- a/core/src/main/java/apoc/meta/Meta.java
+++ b/core/src/main/java/apoc/meta/Meta.java
@@ -83,6 +83,7 @@ public class Meta {
         public boolean existence;
         public String type;
         public boolean array;
+        public List<Object> sample;
         public long left; // 0,1,
         public long right; // 0,1,many
         public List<String> other = new ArrayList<>();
@@ -362,7 +363,7 @@ public class Meta {
         } else {
             throw new IllegalArgumentException("Supported inputs are String, VirtualGraph, Map");
         }
-        return collectMetaData(subGraph, metaConfig).values().stream().flatMap(x -> x.values().stream());
+        return collectMetaData(subGraph, metaConfig.getSampleMetaConfig()).values().stream().flatMap(x -> x.values().stream());
     }
 
     // todo ask index for distinct values if index size < 10 or so
@@ -370,7 +371,7 @@ public class Meta {
     @Procedure("apoc.meta.data")
     @Description("Examines the full graph and returns a table of metadata.")
     public Stream<MetaResult> data(@Name(value = "config",defaultValue = "{}") Map<String,Object> config) {
-        MetaConfig metaConfig = new MetaConfig(config);
+        SampleMetaConfig metaConfig = new SampleMetaConfig(config);
         return collectMetaData(new DatabaseSubGraph(transaction), metaConfig).values().stream().flatMap(x -> x.values().stream());
     }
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/meta/Meta.java
```diff
@@ -909,39 +909,59 @@ private void filterNonExistingRelationships(Map<Pattern, Relationship> vRels, Me
...
-    private boolean relationshipExists(Label labelFromLabel, Label labelToLabel, RelationshipType relationshipType, Direction direction, MetaConfig metaConfig) {
-        Map<String, Long> countStore = getLabelCountStore();
+    /**
+     * relationshipExists uses sampling to check if the relationships added in previous steps exist.
+     * The sample count is the skip count; e.g. if set to 1000 this means every 1000th node will be checked.
+     * A sample count higher than nodeLabel count is equivalent to not checking them.
+     * Note; Each node is still fetched, but the relationships on that node will not be checked
+     * if skipped, which should make it faster.
+     */
+    static boolean relationshipExists(
```
Lojjs, line 932: I wonder if it would be easier to understand if not both this and the previous method had the same name?

core/src/main/java/apoc/meta/Meta.java
```diff
@@ -909,39 +909,59 @@ private void filterNonExistingRelationships(Map<Pattern, Relationship> vRels, Me
...
+    ) {
         try (ResourceIterator<Node> nodes = tx.findNodes(labelFromLabel)) {
-            long count = 1L;
-            String labelName = labelFromLabel.name();
-            long labelCount = countStore.get(labelName);
-            long sample = getSampleForLabelCount(labelCount, metaConfig.getSample());
+            long count = 0L;
+            long labelCount = countStore.get(labelFromLabel.name());
+            // A sample size below 0 means we should check every node.
+            long skipCount = metaConfig.getSample() > 0 ? metaConfig.getSample() : 1;
```
Lojjs, line 945: What if metaConfig.getSample() == 0?

core/src/main/java/apoc/meta/Meta.java
```diff
@@ -909,39 +909,59 @@ private void filterNonExistingRelationships(Map<Pattern, Relationship> vRels, Me
...
+            // Sample size greater than label count is too large, skip
+            // Returning true will keep the relationship, otherwise it may remove all relationships.
+            if (skipCount > labelCount) return true;
```
Lojjs, line 948: I still find it a bit weird that if we have a labelCount = 1000 and a skipCount = 1000 we will check 1 node only and probably filter out most relationships, while if we have skipCount = 1001 it would be a whole other behaviour where we answer true for everything.

An alternative could be to do the count++ last in the loop, then we would check exactly one node (the first one when count is 0) for skipCount >= labelCount. What do you think?

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -201,6 +189,53 @@ public void testMetaGraphMaxRels() {
         db.executeTransactionally("MATCH (n) DETACH DELETE n");
     }
 
+    @Test
+    public void testRelationshipExistsSampling() {
+        List<Long> skipSizes = List.of(2L, 100L, 1000L, -1L, 0L, 1L);
+        List<Long> fromNodeCounts = List.of(4L, 1000L, 100L, 500L, 10000L, 4L);
+        List<Integer> expectedChecks = List.of(2, 10, 0, 500, 10000, 4);
```
Lojjs, line 196: It would be nice if we could also test the impact of maxRels, either in this unit test or in the acceptance tests below

Lojjs (changes requested):
Much better than before, but I have some questions and ideas on how we can improve it further.

core/src/main/java/apoc/meta/Meta.java
```diff
@@ -954,18 +974,30 @@ private void combine(Map<Pair<String, String>, Set<Pattern>> aggregated, Pair<St
...
-        MetaConfig metaConfig = new MetaConfig(config);
-        return metaGraph(new DatabaseSubGraph(transaction), null, null, false, metaConfig);
+    public Stream<GraphResult> graphSample(@Name(value = "config",defaultValue = "{}") @Deprecated Map<String,Object> config) {
+        return metaGraph(
+                new DatabaseSubGraph(transaction),
+                null,
+                null,
+                false,
+                new SampleMetaConfig(null)
+        );
```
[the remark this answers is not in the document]
gem-neo4j: I deprecated the config as it isn't used, then just added a dummy config in, I don't think anywhere wants us to deprecate the actual proc, I just figured there is no point in having a useless config.

core/src/main/java/apoc/meta/Meta.java
gem-neo4j: Most likely! will update

core/src/main/java/apoc/meta/Meta.java
gem-neo4j: Brain slip there hehe woops

core/src/main/java/apoc/meta/Meta.java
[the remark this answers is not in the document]
Lojjs: I was thinking about https://trello.com/c/LFcqUwIy/1918-deprecate-apocmetagraphsample

core/src/test/java/apoc/meta/MetaTest.java
Lojjs: Discussed, there are already such tests

Lojjs (approved):
Nice

PR #307, opened 2023-01-27
[hZww9vSJ] Make the type output of apoc.schema.* procedures more specific

vga91:
Changed `ConstraintDescriptor` to `ConstraintDefinition` in `nodeInfoFromConstraintDefinition`.
This to be consistent with `relationshipInfoFromConstraintDefinition` and to retrieve the constraint type with `constraintDefinition.getConstraintType().name()`, easier to implement and not internal.
The only con is that to get the "userDescriptor" output I have to use this method `ktx.schemaRead().constraintGetForName(constraintDefinition.getName()).userDescription(tokens)`.

Alternatively, I could remain ConstraintDescriptor and derive the type of constraint via a method like this, but it's not future-proof:
```
public static ConstraintType getConstraintType(ConstraintDescriptor descriptor) {
    if (descriptor.isNodeUniquenessConstraint()) return ConstraintType.UNIQUENESS;
    if (descriptor.isNodeKeyConstraint()) return ConstraintType.NODE_KEY;
    if (descriptor.isNodePropertyExistenceConstraint()) return ConstraintType.NODE_PROPERTY_EXISTENCE;
...
    return ConstraintType.UNIQUENESS;
}
```

---
With this implementation, a uniqueness constraint returns 2 results (1 index "RANGE" and a constraint "UNIQUENESS"), consistently to relationships (previously returned a generic "INDEX" result).

I don't think it's a breaking-change but rather a bug as the documentation says `returns ALL indexes and constraints information for all node labels in the database.`.
If in doubt, I can create a card and maybe discuss it together.

---

- changed `indexDescriptor.isUnique() ? "UNIQUENESS" : "INDEX";` to `indexDescriptor.getIndexType().name();`
- removed hardcoded `"INDEX"` from `type` output
- Simplified `Schemas.relationshipInfoFromIndexDescription`

### DIFF
--- a/core/src/test/java/apoc/schema/SchemasTest.java
+++ b/core/src/test/java/apoc/schema/SchemasTest.java
@@ -405,7 +405,7 @@ public class SchemasTest {
 
             assertEquals("RANGE", r.get("type"));
             assertEquals("ONLINE", r.get("status"));
-            final String expectedUserDescBarIdx = "name='constraint_4791de3e', type='RANGE', schema=(:Bar {foo}), indexProvider='range-1.0', owningConstraint=4";
+            final String expectedUserDescBarIdx = "name='constraint_4791de3e', type='RANGE', schema=(:Bar {foo}), indexProvider='range-1.0', owningConstraint";
             Assertions.assertThat(r.get("userDescription").toString()).contains(expectedUserDescBarIdx);
             r = result.next();
 
--- a/core/src/test/java/apoc/schema/SchemasTest.java
+++ b/core/src/test/java/apoc/schema/SchemasTest.java
@@ -849,7 +849,6 @@ public class SchemasTest {
                     result -> {
                         while  (result.hasNext()) {
                             Map<String, Object> r = result.next();
-                            System.out.println("r = " + r);
 
                             assertEquals(1, relConstraints.stream().filter(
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -784,16 +839,17 @@ public void testMultipleUniqueRelationshipConstraints() {
...
                 "ONLINE",
                 "KNOW"
         ));
 
-        testResult(db, "CALL apoc.schema.relationships({})",
+        testResult(db, "CALL apoc.schema.relationships",
                     result -> {
                         while  (result.hasNext()) {
                             Map<String, Object> r = result.next();
+                            System.out.println("r = " + r);
```
gem-neo4j, line 852: Probably best to remove print statements now :)

gem-neo4j (approved):
Looks really good :) Can you please update the docs as well to show the correct type (i.e not uniqueness), just one small comment otherwise! Nice :D 
I also agree this isn't breaking as it is a bug fix!

vga91:
@gem-neo4j I added the docs pr: https://github.com/neo4j/docs-apoc/pull/107
I'll merge this one as soon as it turns green :)

gem-neo4j:
Great, I'll review the docs when I have time, thanks! :D

PR #308, opened 2023-01-30
[4cnpX39y] Bump mockserver-netty version and org.apache.arrow version

gem-neo4j:
Bumps versions to fix vulnerabilities (no CVE present)


AzuObs (approved):
I wonder whether bumping Arrow 3 major versions will contain any backwards incompatibility? I think that even it is does, then this might still be OK to do for security fixes.

If you are able to check this, and you are satisfied with what you see then please consider this approved 😄

gem-neo4j:
I checked through the changelog and couldn't find anything that seemed breaking for us before I did the update :)

PR #309, opened 2023-01-30
[NOID] Update version matrix

gem-neo4j:
Update to latest 4.4.x (and fix typo) and add newer 5.x releases

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -191,9 +191,12 @@ The trailing `<apoc>` part of the version number will be incremented with every 
 [opts=header]
 |===
 |apoc version | neo4j version
+| https://github.com/neo4j/apoc/releases/5.4.1[5.4.1^] | 5.4.0 (5.4.x)
+| https://github.com/neo4j/apoc/releases/5.3.0[5.3.0^] | 5.3.0 (5.3.x)
+| https://github.com/neo4j/apoc/releases/5.2.1[5.2.1^] | 5.2.0 (5.2.x)
 | https://github.com/neo4j/apoc/releases/5.1.0[5.1.0^] | 5.1.0 (5.1.x)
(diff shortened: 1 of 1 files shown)


readme.adoc
```diff
@@ -191,9 +191,12 @@ The trailing `<apoc>` part of the version number will be incremented with every
 [opts=header]
 |===
 |apoc version | neo4j version
+| https://github.com/neo4j/apoc/releases/5.4.1[5.4.1^] | 5.4.0 (5.4.x)
+| https://github.com/neo4j/apoc/releases/5.3.0[5.3.0^] | 5.3.0 (5.3.x)
+| https://github.com/neo4j/apoc/releases/5.2.1[5.2.1^] | 5.2.0 (5.2.x)
 | https://github.com/neo4j/apoc/releases/5.1.0[5.1.0^] | 5.1.0 (5.1.x)
 | https://github.com/neo4j/apoc/releases/5.0.0[5.0.0^] | 5.0.0 (5.0.x)
-| https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/4.4.0.1[4.4.0.1^] | 4.4.0 (4.3.x)
+| https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/4.4.0.12[4.4.0.12^] | 4.4.0 (4.4.x)
```
ncordon, line 199: Do we want to keep maintaining this manually? For 4.4 it was unclear which versions we should use, but for 5.x, it should be the highest 5.x.y, so what's your opinion on removing the version matrix and including that phrasing?

gem-neo4j: That's fair, I just did this now because it was speedy to do, but in the long run it will be slower! I'll leave extended the old way as the extended folks have been maintaining it, but for us I think it makes sense to remove :) I will update it!

PR #310, opened 2023-01-30
[AJmycukR] Fix for apoc.import.graphml

Lojjs:


core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
```diff
@@ -200,6 +200,8 @@ public long parseXML(Reader input) throws XMLStreamException {
         XMLInputFactory inputFactory = XMLInputFactory.newInstance();
         inputFactory.setProperty("javax.xml.stream.isCoalescing", true);
         inputFactory.setProperty(XMLInputFactory.IS_NAMESPACE_AWARE, true);
+        inputFactory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
+        inputFactory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);
```
Lojjs, line 204: Another possible approach could be to try to reuse the same XML parser that we use for the apoc.xml procedures, not sure how feasible that would be

AzuObs: I think that different XML parsers in Java have different behaviour and work well for different things. The XMLInputFactory is however used in another place IIRC. 
 
 We could have our own XMLParserFactory which returns an XMLEventReader with sensible defaults. I leave it up to you whether you'd like to explore that. However, neither do I see this procedure staying in APOC Core in the long term, nor do I anticipate many code changes in this procedure between now and 2025 when 4.4 goes out of support.

AzuObs (approved):
Nice job 🙂

PR #313, opened 2023-02-01
[HQABUD5U] Improve coverNodes speed

gem-neo4j:
Use a hashset instead to improve performance of contains for larger node lists.


AzuObs (approved):
Nice improvement 🙂

PR #321, opened 2023-02-07
[NOID] Cleans up resources after the tests

ncordon:
## What
Cleans up resources after the tests.

Also removes a couple of tests that were not testing anything related to apoc: `ArrayListTest` and `EqualityTest`

## Why
Because we were leaving open many `DbmsRule`s, which can leave lots of background threads without closing, causing the CI, as we've seen, to go red oftentimes.


gem-neo4j (approved):
Looks good! Thanks for this :D Hopefully it helps 🙌

PR #322, opened 2023-02-08
[8n02RETv] Fixes the cluster setup for the trigger tests

ncordon:
## What

Solves some cluster tests flakiness.

The setup we had was to wait for every instance to return 200 in the `db/neo4j/cluster/available` endpoint. But in 5.x neo4j is not allocated by default in every instance. This PR forces it to be allocated in every instance and reshapes the trigger tests so they test for what they are supposed to. This is a temporary hack, we should revisit the setup, because we never reshaped it from 4.x (causal clusters) to 5.x (autonomoous cluster). 

## Why
The cluster tests were broken in general and I had realized because there was some flakiness in the CI related to these tests. An instance was always succeeding to start, the one with neo4j allocated in it and if that instance happened to have a be a system leader as well, it was failing for the drop trigger (because we cannot drop something we have not created and drop would return empty results rather than something).

### DIFF
--- a/core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
+++ b/core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
@@ -22,6 +22,7 @@ import static apoc.trigger.Trigger.SYS_DB_NON_WRITER_ERROR;
 import static apoc.trigger.TriggerNewProcedures.TRIGGER_NOT_ROUTED_ERROR;
 import static apoc.util.TestContainerUtil.testCall;
 import static apoc.util.TestContainerUtil.testResult;
+import static org.junit.Assert.assertThrows;
 import static org.neo4j.configuration.GraphDatabaseSettings.DEFAULT_DATABASE_NAME;
 import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME;
 import static org.junit.Assert.assertEquals;
@@ -152,14 +153,11 @@ public class TriggerClusterRoutingTest {
             Session session = getSessionForDb(instance, dbName);
 
             if (!dbIsWriter(SYSTEM_DATABASE_NAME, session, getBoltAddress(instance))) {
-                try {
-                    testCall(session, triggerOperation,
+                Exception e = assertThrows(Exception.class, () -> testCall(session, triggerOperation,
                             Map.of("name", triggerName),
-                            row -> fail("Should fail because of non writer trigger addition"));
-                } catch (Exception e) {
-                    String errorMsg = e.getMessage();
-                    assertTrue("The actual message is: " + errorMsg, errorMsg.contains(expectedError));
-                }
+                            row -> fail("Should fail because of non writer trigger addition")));
+                String errorMsg = e.getMessage();
+                assertTrue("The actual message is: " + errorMsg, errorMsg.contains(expectedError));
             }
         }
     }


common/build.gradle
```diff
@@ -56,7 +56,7 @@ dependencies {
     api group: 'org.hdrhistogram', name: 'HdrHistogram', version: '2.1.9'
     api group: 'org.apache.commons', name: 'commons-collections4', version: '4.2'
     // We need this to avoid seeing SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder" on startup
-    api group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.36'
+    api group: 'org.slf4j', name: 'slf4j-simple', version: '2.0.0'
```
ncordon, line 59: Without this change the cluster containers complain the logger is too old and they don't log anything

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -31,7 +31,7 @@ public class TriggerClusterRoutingTest {
     public static void setupCluster() {
         cluster = TestContainerUtil
                 .createEnterpriseCluster( List.of(TestContainerUtil.ApocPackage.CORE), 
-                        3, 1, 
+                        3, 0,
```
ncordon, line 34: Don't allocate a read replica. We weren't testing anything with the read replica anyway (because we were skipping it inside the tests) and read replicas have a different behaviour in 5.x anyway, so it was failing to start up

core/build.gradle
```diff
@@ -88,7 +88,9 @@ dependencies {
     testImplementation group: 'junit', name: 'junit', version: '4.13.2'
     testImplementation group: 'com.github.stefanbirkner', name: 'system-rules', version: '1.19.0'
     testImplementation group: 'com.amazonaws', name: 'aws-java-sdk-s3', version: '1.12.348'
-    testImplementation group: 'org.mock-server', name: 'mockserver-netty', version: '5.15.0'
+    testImplementation group: 'org.mock-server', name: 'mockserver-netty', version: '5.15.0', {
+        exclude group: 'org.slf4j', module: 'slf4j-api'
```
ncordon, line 92: Without this the container logs don't work because we have a conflicting sl4j-api version (2.x.y is coming from here, 1.7.36 is coming from :common)

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -49,65 +49,82 @@ public static void bringDownCluster() {
...
     public void testTriggerRemoveAllowedOnlyInSysLeaderMember() {
-        final String query = "CALL apoc.trigger.remove($name)";
-        triggerInSysLeaderMemberCommon(query, SYS_DB_NON_WRITER_ERROR, DEFAULT_DATABASE_NAME);
+        final String addTrigger = "CALL apoc.trigger.add($name, 'RETURN 1', {})";
+        final String removeTrigger = "CALL apoc.trigger.remove($name)";
+        String triggerName = randomTriggerName();
+
+        succeedsInLeader(addTrigger, triggerName, DEFAULT_DATABASE_NAME);
+        succeedsInLeader(removeTrigger, triggerName, DEFAULT_DATABASE_NAME);
+        failsInFollowers(removeTrigger, triggerName, SYS_DB_NON_WRITER_ERROR, DEFAULT_DATABASE_NAME);
```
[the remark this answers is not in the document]
gem-neo4j: I guess it doesn't matter that it tries to remove an already removed trigger in the followers?

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -51,87 +57,130 @@ public static void bringDownCluster() {
...
-    private static void triggerInSysLeaderMemberCommon(String query, String triggerNotRoutedError, String dbName) {
-        final BiConsumer<Session, String> testTrigger = (session, name) -> testCall(session, query,
-                Map.of("name", name),
-                row -> assertEquals(name, row.get("name")));
-        triggerInSysLeaderMemberCommon(query, triggerNotRoutedError, dbName, false, testTrigger);
+    private static void succeedsInLeader(String triggerOperation, String triggerName, String dbName) {
+        for (Neo4jContainerExtension instance: clusterMembers) {
+            Session session = getSessionForDb(instance, dbName);
+
+            if (dbIsWriter(SYSTEM_DATABASE_NAME, session, getBoltAddress(instance))) {
```
gem-neo4j, line 122: So if this doesn't evaluate to true, we will not actually test the trigger operation succeeded in leader? I assume there will always be a writer though? :)

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
```diff
@@ -51,87 +57,130 @@ public static void bringDownCluster() {
...
+
+    private static void failsInFollowers(String triggerOperation, String triggerName, String expectedError, String dbName) {
+        for (Neo4jContainerExtension instance: clusterMembers) {
+            Session session = getSessionForDb(instance, dbName);
+
+            if (!dbIsWriter(SYSTEM_DATABASE_NAME, session, getBoltAddress(instance))) {
                 try {
-                    testCall(session, query,
-                            Map.of("name", UUID.randomUUID().toString()),
+                    testCall(session, triggerOperation,
```
gem-neo4j, line 156: There is an  Assert.assertThrows() function that we could use potentially :)

gem-neo4j (approved):
LGTM, just a few comments, nothing major :) Looks like a nice tidy up as well!

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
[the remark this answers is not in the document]
ncordon: Nope, it would fail before actually doing that

core/src/test/java/apoc/trigger/TriggerClusterRoutingTest.java
ncordon: There must always be a leader (=writer) for the system database among the instances of the cluster, so we should always be testing something here

PR #326, opened 2023-02-09
[F61iNGQX] Splits integration tests into their own package

ncordon:
## What
* Splits integration tests into a separate package `it`.
* Executes the projects in parallel (`org.gradle.parallel=true` in `gradle.properties` file).
* Chunks the CI into compile + each of subprojects tests vs just a job doing compilation + tests.

## Why
Because:
* That way we can actually run the tests in parallel.
* We halved the time of execution for GitHub Actions, TeamCity and local testing.
* We can detect what project was having test failures better, or wether the failure happened in compilation instead.

## GitHub Actions
The CI projects have been split and refactored, we are down from ~30 min to ~16 min

## TeamCity
By using gradle parallelism we are down from ~30 min to 16min as well

## Locally
Because I have more CPU processors in my laptop than the CI (2 in GitHub Actions, for example), now you can run core:tests (used to take ~30 min locally) in roughly 5 min. The it-tests is a different story, because they require lots of docker containers, so they are still very slow and we cannot parallelise them because some of the schema tests there would break (already tried that).

Screenshot 2023-02-10 at 14 25 35


.github/workflows/CI.yaml
```diff
@@ -10,65 +10,90 @@ env:
...
-      DOCKER_ENTERPRISE_DEV_URL: ${{ secrets.DOCKER_ENTERPRISE_DEV_URL }}
-      DOCKER_COMMUNITY_DEV_URL: ${{ secrets.DOCKER_COMMUNITY_DEV_URL }}
-      ENTERPRISE_TAR: enterprise-docker.tar
-      COMMUNITY_TAR: community-docker.tar
-
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-
+      - uses: actions/checkout@v3
```
ncordon, line 34: Bumped all the versions, got rid of the warnings we had executing the jobs before:

core/src/test/java/apoc/StartupTest.java
```diff
@@ -1,48 +0,0 @@
...
-            neo4jContainer.start();
-
-            Session session = neo4jContainer.getSession();
-            int procedureCount = session.run("SHOW PROCEDURES YIELD name WHERE name STARTS WITH 'apoc' RETURN count(*) AS count").peek().get("count").asInt();
-            int functionCount = session.run("SHOW FUNCTIONS YIELD name WHERE name STARTS WITH 'apoc' RETURN count(*) AS count").peek().get("count").asInt();
-            int coreCount = session.run("CALL apoc.help('') YIELD core WHERE core = true RETURN count(*) AS count").peek().get("count").asInt();
-
-            assertTrue(procedureCount > 0);
-            assertTrue(functionCount > 0);
-            assertTrue(coreCount > 0);
```
ncordon, line 37: I got rid of this test because there's another `StartupTest` and they seem to do the same

build.gradle
```diff
@@ -75,7 +75,7 @@ subprojects {
                 'neo4jCommunityDockerImage': System.getProperty("NEO4JVERSION") ? 'neo4j:' + System.getProperty("NEO4JVERSION") : 'neo4j:5.6.0',
                 'coreDir': 'core'
 
-        maxHeapSize = "8G"
+        maxHeapSize = "5G"
```
ncordon, line 78: I'm not sure this had any impact, but GitHub Actions machines have 6GB of memory, so if this was honoured it could be problematic

build.gradle
```diff
@@ -108,7 +108,7 @@ subprojects {
         }
 
         testLogging.showStandardStreams = true
-        testLogging.exceptionFormat = 'full'
+        testLogging.exceptionFormat = 'short'
```
ncordon, line 111: Try to print shorter stack traces in the tests

core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -166,6 +167,8 @@ public void testSlottedRuntime() {
         assertFalse(Periodic.slottedRuntime("cypher expressionEngine=compiled MATCH (n) RETURN n").contains(" runtime=slotted cypher"));
 
     }
+
+    @Ignore
```
ncordon, line 171: This is flaky, I've carded this

.github/workflows/CI.yaml
```diff
@@ -10,65 +10,90 @@ env:
...
+
+  common-tests:
+    runs-on: ubuntu-latest
+    needs: compile
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run common tests
+        uses: ./.github/actions/test-gradle-project
+        with:
+          project-name: common
```
vga91, line 54:
```suggestion
    env:
      PROJECT_NAME: "core"
    steps:
      - uses: actions/checkout@v3
      - name: Run ${{env.PROJECT_NAME}} tests
        uses: ./.github/actions/test-gradle-project
        with:
          project-name: $PROJECT_NAME
```
Personally I would write it like this, i.e. the project name as a variable,
so as to avoid possible discrepancies between `name:`, `project-name`, and any lines that may be inserted in the future.

 But it's subjective, it's up to you if you want to change it or not.

vga91 (commented):
Really nice work!

I would suggest just a few small changes.

Also I'm wondering, how come some it-test schemas don't work in parallel? 
It would be nice to parallelize them, if feasible, since as you said, they are very slow.

build.gradle
vga91: By doing a `println(testLogging.exceptionFormat)` it seems to me that this format is already 'short',
so I guess we can directly remove the line, if I'm not missing something

build.gradle
ncordon: 512m to run the tests is too little. The settings in 4.4 should be changed, not these ones

core/build.gradle
```diff
@@ -45,6 +45,10 @@ generateGrammarSource {
     arguments += ["-package", "apoc.custom"]
 }
 
+test {
+    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+}
```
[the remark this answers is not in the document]
ncordon: Good catch!

.github/workflows/CI.yaml
ncordon: Good point. I've simplified the repeated code with a matrix.

it/build.gradle
```diff
@@ -1,8 +1,11 @@
-description = 'APOC :: Test Startup Module'
+description = 'APOC :: Integration Tests Module'
 
 dependencies {
+    testImplementation project(':common')
     testImplementation project(':core')
     testImplementation project(':test-utils')
+    testImplementation project(":common").sourceSets.test.output
```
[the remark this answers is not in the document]
ncordon: We don't directly for the code. We do for the test resources:

build.gradle
ncordon: Ah now I get your point. Don't you think anyway having a consistent setup for local and CI runs (like having the same heap space) could be better?

vga91: Yeah well, thinking about it maybe it's not the best to have 2 different setups.
That's okay then :)

vga91 (approved):
LGTM, really great job!

PR #328, opened 2023-02-15
[iTwFvqQ2] apoc.create.clonePathsToVirtual() returns more relationships than it should

vga91:
Added cache for `apoc.create.clonePathsToVirtual` in order to reuse the same virtual rel if already present,
so we don't visualize multiple relationships

### DIFF
--- a/core/src/main/java/apoc/create/Create.java
+++ b/core/src/main/java/apoc/create/Create.java
@@ -8,6 +8,7 @@ import apoc.uuid.UuidUtil;
 import org.neo4j.graphdb.*;
 import org.neo4j.procedure.*;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -315,6 +318,51 @@ public void testClonePathShouldNotEditOriginalOne() {
                     assertFalse(end.hasProperty("brazorf"));
                 });
     }
+
+    @Test
+    public void testClonePathShouldNotDuplicateRelsWithMultipaths() {
+        //create path with single rels
+        db.executeTransactionally("""
+                          CREATE (n:Node{id:0}),
```
nadja-muller, line 326: I would simplify the test as much possible to:

```
CREATE (n1:Node),
              (n2:Node),
              (n3:Node),
              (n1)-[:R]->(n2),
              (n1)-[:R]->(n2)-[:R]->(n3)
```

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -315,6 +318,51 @@ public void testClonePathShouldNotEditOriginalOne() {
...
+                            (n)-[:PARENT]->(n2),
+                            (n)-[:PARENT]->(n3),
+                            (n)-[:PARENT]->(n5),
+                            (n)-[:PARENT]->(n6)-[:PARENT]->(n7),
+                            (n)-[:PARENT]->(n4)-[:PARENT]->(n7)-[:PARENT]->(n2)-[:PARENT]->(n3)"""
+        );
+        
+        // returns a list with all rels
+        testCall(db, """
+                MATCH p=(n:Node{id: 0})-[:PARENT*..4]->(v:Node)
```
nadja-muller, line 342: With:
`MATCH p=()-[:R*..2]->() 
WITH collect(p) AS paths
CALL apoc.create.clonePathsToVirtual(paths)
YIELD path
RETURN path`

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -315,6 +318,51 @@ public void testClonePathShouldNotEditOriginalOne() {
...
+                RETURN apoc.coll.flatten(pathRels) as rels""", r -> {
+            final List<Relationship> paths = (List<Relationship>) r.get("rels");
+            assertFalse(paths.isEmpty());
+
+            // group the rels by the pair of start-end node 
+            final Map<Pair<String, String>, List<Relationship>> collect = paths.stream()
+                    .collect(Collectors.groupingBy(i -> Pair.of(
+                            i.getStartNode().getElementId(),
+                            i.getEndNode().getElementId()))
+                    );
```
nadja-muller, line 356: The verification seems overly complicated to me. With the smaller sample set above, it would be enough to verify that:
* The list of relationship has size 3.
* The collection `Map<Relationship, Long> collection = rels.stream().collect(Collectors.groupingBy(/*Pair*/) has size 2.

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -315,6 +318,51 @@ public void testClonePathShouldNotEditOriginalOne() {
...
+        
+        // returns a list with all rels
+        testCall(db, """
+                MATCH p=(n:Node{id: 0})-[:PARENT*..4]->(v:Node)
+                WITH collect(p) as p
+                CALL apoc.create.clonePathsToVirtual(p)
+                YIELD path
+                WITH collect( relationships(path) ) as pathRels
+                RETURN apoc.coll.flatten(pathRels) as rels""", r -> {
+            final List<Relationship> paths = (List<Relationship>) r.get("rels");
```
nadja-muller, line 348: I think `paths` is a bit of a confusing name here, as it is a list of relationships. How about `rels` or `relationships `instead?

nadja-muller (changes requested):
The fix looks good. It would be nice to simplify the test a bit.

core/src/main/java/apoc/create/Create.java
```diff
@@ -196,27 +197,46 @@ public Stream<VirtualPathResult> virtualPath(@Name("labelsN") List<String> label
     @Procedure("apoc.create.clonePathToVirtual")
     @Description("Takes the given path and returns a virtual representation of it.")
     public Stream<PathResult> clonePathToVirtual(@Name("path") Path path) {
-        return Stream.of(createVirtualPath(path));
+        // given that it accepts a single path as a parameter the cache is not necessary
```
nadja-muller, line 200: Also, I wouldn't call it the map a cache, since you use it for tracking the relationships already created and not primarily as a performance optimization. Perhaps we could call it a "relationshipMap" or "createdRelationships"?

vga91: Changed to `relationshipMap`

core/src/test/java/apoc/create/CreateTest.java
vga91: Changed. 
I'm not really sure if that's the change you meant, because if I'm not mistaken `Map<Relationship, Long>` and `.collect(Collectors.groupingBy(/Pair/)` are incorrect. 
Let me know if as changed is what you wanted, in case I can re-change it :)

nadja-muller (approved):

PR #334, opened 2023-02-20
[F61iNGQX] Removes parallel execution in TeamCity

ncordon:
Keeps it in GitHub Actions.

## Why
Because TeamCity seems to not be very happy with it at the minute. Probably because all projects get executed at the same time, rather than the parallelism been intra-project, like in GitHub Actions


gem-neo4j (approved):
LGTM :)

PR #335, opened 2023-02-21
[NOID] Fix error message for apoc.schema.nodes/relationships with include and exclude confs both valuated

vga91:
Since the config keys are case sensitive, these error messages are wrong:
```
Parameters relationships and excluderelationships are both valuated. 

Parameters labels and excludelabels are both valuated.
```


Lojjs (approved):
Looks good, thanks for fixing!

PR #337, opened 2023-02-28
[F61iNGQX] Executes integration tests sequentially

ncordon:
## Why
Because these tests spin up a lot of docker containers, especially in TeamCity, which can lead to flakiness


gem-neo4j (approved):
LGTM :)

PR #338, opened 2023-03-01
[F61iNGQX] Moves S3 tests to integration tests package

ncordon:
## Why
Because they seem to be flaky as well, they depend on an S3 container, so it's better to have them in the integration tests package.


gem-neo4j (approved):
LGTM :) Thanks for all the test cleanup 😄

PR #339, opened 2023-03-03
[NOID] Bump version to 5.7.0

gem-neo4j:


Lojjs (approved):
You need to restart the build when Neo4j version has been updated, but otherwise looks good

PR #343, opened 2023-03-07
[Bc2lkk3N] Fixed testImportCsvTerminate and added TerminationGuard to apoc.import.csv

vga91:
- Added `TerminationGuard` to `apoc.import.csv`

---

The test didn't fail because it produced a false positive, 
since without the TerminationGuard, the procedure complete the whole procedure when the `TERMINATE TRANSACTION <procTransactionId>` is called, but despite this an `"Explicitly terminated by the user."` is eventually returned.

Therefore, with procedure duration less than the set `db.transaction.timeout`, 
both with and without the TerminationGuard, the test was green.
Apparently I didn't replicate it because, unlike gh actions and locally, 
with teamcity sometimes the procedure lasts more then `timeout` time.

So, I added an additional check to `checkTerminationGuard`, 
to verify that the procedure actually lasts longer than the timeout 
and I weighted the `testImportCsvTerminate`.

---

Not strictly card-related (to make the pr green):
- Added `TerminationGuard` to  `apoc.refactor.cloneSubgraph`

### DIFF
--- a/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
+++ b/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
@@ -12,6 +12,7 @@ import com.opencsv.CSVReader;
 import com.opencsv.CSVReaderBuilder;
 import org.neo4j.graphdb.*;
 import org.neo4j.logging.Log;
+import org.neo4j.procedure.TerminationGuard;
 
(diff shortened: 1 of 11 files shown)


nadja-muller (approved):
Looks reasonable to me

PR #344, opened 2023-03-07
[h99LDu82] Fix testRunTimeboxedWithTerminationInnerTransaction

vga91:
It seems to me that the test fails (because the procedure returns empty value) only sometimes with the embedded database and only with neo4j >5.6.0.

Unfortunately I don't quite understand why,
so for now I've added a `hasNext()` to the test which seems to work

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -69,7 +69,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: ['common', 'core', 'processor', 'test-utils', 'it']
+        project: ['core']
     env:
       DOCKER_ENTERPRISE_DEV_URL: ${{ secrets.DOCKER_ENTERPRISE_DEV_URL }}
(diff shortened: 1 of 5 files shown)


gem-neo4j:
Hey! I tried out the test query on my local 5.5 db and was able to get both the result of {"0": 0} and empty. I think the bug here is definitely more in the procedure than the test, and will make a card for us to fix the procedure rather than allow the test to hide the buggy behaviour :)

vga91:
@gem-neo4j Since the flakiness was likely due to [this thing](https://github.com/neo4j/apoc/pull/373), 
I guess we can close this pr now and verify that the error is no longer present. 
Wdyt?

gem-neo4j:
> @gem-neo4j Since the flakiness was likely due to [this thing](https://github.com/neo4j/apoc/pull/373), I guess we can close this pr now and verify that the error is no longer present. Wdyt?

Yes, sounds good :)

PR #346, opened 2023-03-09
[H5vz3qhy] Prevent cascade trigger transactions

vga91:
Added a `CALL tx.setMetaData(<TRIGGER_META>)` procedure in transactions with phase `after` and `afterAsync`,
in order to prevent cascade trigger transactions coming from `TriggerHandler` itself,
so that we can avoid some deadlock conditions.

---

I don't know if it is somehow directly related to this pr,
but I had to change the `testRemoveNode` too, 
because it sometimes failed as the previous `CREATE (f:Foo)` had id = 0.

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerHandler.java
+++ b/core/src/main/java/apoc/trigger/TriggerHandler.java
@@ -35,7 +35,7 @@ import static apoc.ApocConfig.apocConfig;
 public class TriggerHandler extends LifecycleAdapter implements TransactionEventListener<Void> {
 
     private enum Phase {before, after, rollback, afterAsync}
-    private static final Map<String, Object> TRIGGER_META = Map.of("trigger", true);
+    private static final Map<String, Object> TRIGGER_META = Map.of("apoc.trigger", true);
 
     public static final String TRIGGER_REFRESH = "apoc.trigger.refresh";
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -35,6 +35,7 @@
 public class TriggerHandler extends LifecycleAdapter implements TransactionEventListener<Void> {
 
     private enum Phase {before, after, rollback, afterAsync}
+    private static final Map<String, Object> TRIGGER_META = Map.of("trigger", true);
```
ncordon, line 38: `Map.of("apoc.trigger", true)` would be better in my opinion. Just in case people start seeing things in their transactions they don't know where they came from, that's more self explanatory than just trigger

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -203,8 +204,17 @@ public Void beforeCommit(TransactionData txData, Transaction transaction, GraphD
 
     @Override
     public void afterCommit(TransactionData txData, Void state, GraphDatabaseService databaseService) {
+        // if `txData.metaData()` is equal to TRIGGER_META, it means that the transaction
+        // if it is equal, it means that the transaction comes from another TriggerHandler transaction,
+        // therefore the execution must be blocked to prevent a deadlock due to cascading transactions
```
ncordon, line 209:
```suggestion
        // if `txData.metaData()` is equal to TRIGGER_META, it means that the transaction
        // comes from another TriggerHandler transaction,
        // therefore the execution must be blocked to prevent a deadlock due to cascading transactions
```

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -203,8 +204,17 @@ public Void beforeCommit(TransactionData txData, Transaction transaction, GraphD
...
+        Map<String, Object> metaData = txData.metaData();
+        if (metaData.equals(TRIGGER_META)) {
```
ncordon, line 211: This is personal preference but if you extract this block outside to a function:

```
public Boolean isTransactionCreatedByTrigger() {
  Map<String, Object> metaData = txData.metaData();
  return metaData.equals(TRIGGER_META);
}
```

then this method becomes more self-documented

ncordon (approved):

PR #347, opened 2023-03-13
[TOQo1n9M] Core part - apoc-hadoop dependency is conflicting 5.x (neo4j-contrib/neo4j-apoc-procedures#3450)

vga91:
Cherry-pick (core part) of https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3450

Extended part: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3454.

- Because of split, now we have 2 equivalent `apoc.ApocSignatures` which collect the list of procedures and functions, 
one for core and one for extended, so I added a `String projectPath` in order to create 2 different classes (`ApocSignaturesCore` and `ApocSignaturesExtended`).
- Created a `copyFilesToPlugin(..)` to be reused in Extended
- Changed baseDir, pluginsFolder and extendedDir to public, reusable in Ext. too

### DIFF
--- a/it/src/test/java/apoc/it/core/StartupTest.java
+++ b/it/src/test/java/apoc/it/core/StartupTest.java
@@ -1,5 +1,5 @@
 package apoc.it.core;
-import apoc.ApocSignaturesCore;
+import apoc.ApocSignatures;
 import apoc.util.Neo4jContainerExtension;
 import apoc.util.TestContainerUtil;
 import apoc.util.TestUtil;
@@ -75,8 +75,8 @@ public class StartupTest {
                             .list(record -> record.get("name").asString());
 
 
-                    assertEquals(sorted(ApocSignaturesCore.PROCEDURES), procedureNames);
-                    assertEquals(sorted(ApocSignaturesCore.FUNCTIONS), functionNames);
+                    assertEquals(sorted(ApocSignatures.PROCEDURES), procedureNames);
+                    assertEquals(sorted(ApocSignatures.FUNCTIONS), functionNames);
                 }
 
                 neo4jContainer.close();
--- a/it/src/test/java/apoc/it/core/StartupTest.java
+++ b/it/src/test/java/apoc/it/core/StartupTest.java
@@ -1,5 +1,5 @@
 package apoc.it.core;
-import apoc.ApocSignatures;
+import apoc.ApocSignaturesCore;
 import apoc.util.Neo4jContainerExtension;
 import apoc.util.TestContainerUtil;
 import apoc.util.TestUtil;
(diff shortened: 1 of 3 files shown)


vga91:
@lojjs I know it's a cherry pick of a pr you already approved, 
but due to the split in 5.x I had to make a few changes from the original one.

Do you (or someone else) want to take a look here as well, by any chance?

processor/src/main/java/apoc/processor/ExtensionClassWriter.java
```diff
@@ -25,16 +29,37 @@ public void write(List<String> procedureSignatures,
...
         }
     }
 
-    private TypeSpec defineClass(List<String> procedureSignatures, List<String> userFunctionSignatures) {
-        return TypeSpec.classBuilder("ApocSignatures")
+    private String getProjectPath() throws IOException {
+        // create and delete a file to retrieve the current project (`core` or `extended`)
+        FileObject resource = filer.createResource(StandardLocation.SOURCE_OUTPUT, "", "tmp", (Element[]) null);
+        String projectPath = resource.getName();
+        resource.delete();
```
Lojjs, line 47: This feel a bit over complicated to me. Cannot we not e.g. send in a boolean to the write method depending on if we call it from core or extended?

it/src/test/java/apoc/it/core/StartupTest.java
```diff
@@ -75,8 +75,8 @@ public void compare_with_sources() {
                             .list(record -> record.get("name").asString());
 
 
-                    assertEquals(sorted(ApocSignatures.PROCEDURES), procedureNames);
-                    assertEquals(sorted(ApocSignatures.FUNCTIONS), functionNames);
+                    assertEquals(sorted(ApocSignaturesCore.PROCEDURES), procedureNames);
```
Lojjs, line 78: I had a brief look at your extended PR too. How come we test both core and extended in extended PR? I would (maybe naively) have thought that it would make more sense to have the checkCoreWithExtraDependenciesJars test here and the other 2 tests in the extended repo?

Lojjs (changes requested):
I have a couple of concerns, not sure if my ideas work though

test-utils/src/main/java/apoc/util/TestContainerUtil.java
```diff
@@ -46,9 +47,10 @@ public enum ApocPackage {
 
     private TestContainerUtil() {}
 
-    private static File baseDir = Paths.get("..").toFile();
+    public static File baseDir = Paths.get("..").toFile();
+    public static File pluginsFolder = new File(baseDir, "build/plugins");
```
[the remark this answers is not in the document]
vga91: No, just a mistake:X. Removed from line 89

processor/src/main/java/apoc/processor/ExtensionClassWriter.java
vga91: Changed with a boolean

it/src/test/java/apoc/it/core/StartupTest.java
Lojjs: Ah ok. I don't think I actually have tried to build them myself, so you might be right about that. If it was a conscious decision, I think it is ok.

processor/src/main/java/apoc/processor/ExtensionClassWriter.java
Lojjs: I was more thinking about if we could avoid creating a temporary file, but looking closer I think it might not be so easy to do it in a smarter way. With your change it is at least a bit cleaner :)

Lojjs (approved):
With your new changes, I think this is good enough

PR #348, opened 2023-03-14
[madSHaLz] APOC triggers aren't updated after a user deletes a database

vga91:
Implemented 3rd solution - `DatabaseEventListener`,
so that we can delete every system node belonging to the database that we delete.

Added a `cleanUpSystemNodes()` for backward compatibility

### DIFF
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -1,35 +1,56 @@
 package apoc.cypher;
 
 import apoc.ApocConfig;
+import apoc.SystemLabels;
 import apoc.util.Util;
 import apoc.util.collection.Iterators;
 import apoc.version.Version;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.lang3.StringUtils;
 import org.neo4j.common.DependencyResolver;
+import org.neo4j.configuration.GraphDatabaseSettings;
+import org.neo4j.dbms.api.DatabaseManagementService;
+import org.neo4j.graphdb.Label;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.event.DatabaseEventContext;
+import org.neo4j.graphdb.event.DatabaseEventListener;
+import org.neo4j.kernel.api.procedure.GlobalProcedures;
 import org.neo4j.kernel.availability.AvailabilityListener;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.kernel.monitoring.DatabaseEventListeners;
 import org.neo4j.logging.Log;
 
 import java.util.Collection;
 
+import java.util.ConcurrentModificationException;
+import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.function.BiConsumer;
+
+import static apoc.SystemPropertyKeys.database;
 
 public class CypherInitializer implements AvailabilityListener {
 
     private final GraphDatabaseAPI db;
     private final Log userLog;
     private final DependencyResolver dependencyResolver;
+    private final DatabaseManagementService databaseManagementService;
+    private final DatabaseEventListeners databaseEventListeners;
 
     /**
      * indicates the status of the initializer, to be used for tests to ensure initializer operations are already done
      */
     private volatile boolean finished = false;
 
-    public CypherInitializer(GraphDatabaseAPI db, Log userLog) {
+    public CypherInitializer(GraphDatabaseAPI db, Log userLog,
+                             DatabaseManagementService databaseManagementService,
+                             DatabaseEventListeners databaseEventListeners) {
         this.db = db;
         this.userLog = userLog;
+        this.databaseManagementService = databaseManagementService;
+        this.databaseEventListeners = databaseEventListeners;
         this.dependencyResolver = db.getDependencyResolver();
     }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -70,6 +88,15 @@ public void available() {
                     } catch (Exception ignored) {
                         userLog.info("Cannot check APOC version compatibility because of a transient error. Retrying your request at a later time may succeed");
                     }
+
+                    // this block is already in system db, so we can directly execute system operation
+                    // for backward compatibility: clear system nodes at start-up if the corresponding database name doesn't exist
+                    // placed here instead of e.g. `ApocConfig` because at this point we are sure that all databases are recovered
+                    cleanUpSystemNodes();
```
ncordon, line 95: I don't think this is true. Databases are brought up in an eventual consistent way in 5.x so when system is up neo4j could not be up

it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
```diff
@@ -46,10 +49,15 @@ public class TriggerEnterpriseFeaturesTest {
...
+        final String cypherInitializer = String.format("%s.%s.0",
+                APOC_CONFIG_INITIALIZER, SYSTEM_DATABASE_NAME);
+        final String createInitDb = String.format("CREATE DATABASE %s IF NOT EXISTS", INIT_DB);
+
         // We build the project, the artifact will be placed into ./build/libs
         neo4jContainer = createEnterpriseDB(List.of(TestContainerUtil.ApocPackage.CORE), true)
                 .withEnv(APOC_TRIGGER_ENABLED, "true")
-                .withEnv(TRIGGER_REFRESH, String.valueOf(TRIGGER_DEFAULT_REFRESH));
+                .withEnv(TRIGGER_REFRESH, String.valueOf(TRIGGER_DEFAULT_REFRESH))
+                .withEnv(cypherInitializer, createInitDb);
```
ncordon, line 60: Why do we need this?

core/src/test/java/apoc/trigger/TriggerRestartTest.java
```diff
@@ -62,6 +74,49 @@ private void restartDb() {
...
+        restartDb();
+
+        withSystemDb(tx -> {
+            Iterator<Node> dbNotExistentNodes = tx.findNodes(ApocTrigger, database.name(), dbName);
+            assertFalse( dbNotExistentNodes.hasNext() );
+
+            Iterator<Node> neo4jNodes = tx.findNodes(ApocTrigger, database.name(), db.databaseName());
+            assertTrue( neo4jNodes.hasNext() );
+        });
+    }
```
ncordon, line 111: I think we should add a test where we install a trigger for neo4j, drop neo4j and eventually the trigger nodes are wiped out

vga91: I added it in `TriggerEnterpriseFeaturesTest.testDeleteTriggersAfterDatabaseDeletion`

it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
vga91: Sorry, I had put some extra code that I probably mistakenly deleted after a rebase :X
It's to check that even if I create a db via `apoc.initializer.*` and then delete it, the implementation still works.
See the test `TriggerEnterpriseFeaturesTest.testDeleteTriggersAfterDatabaseDeletion` (line 193)

core/src/main/java/apoc/cypher/CypherInitializer.java
vga91: Yes indeed it is not safe.
I've been looking for other ways to do this but I don't think there is currently a method that isn't completely error prone, but maybe I'm missing something.

Perhaps it's better not to do this thing at all and document somewhere that for backward compatibility we have to delete the triggers manually. Wdyt?

core/src/test/java/apoc/trigger/TriggerRestartTest.java
```diff
@@ -10,19 +13,28 @@
...
 import static apoc.ApocConfig.SUN_JAVA_COMMAND;
+import static apoc.SystemLabels.ApocTrigger;
+import static apoc.SystemPropertyKeys.database;
+import static apoc.SystemPropertyKeys.name;
 import static apoc.trigger.TriggerTestUtil.TRIGGER_DEFAULT_REFRESH;
 import static apoc.trigger.TriggerTestUtil.awaitTriggerDiscovered;
 import static apoc.util.TestUtil.waitDbsAvailable;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
ncordon, line 37: I think these imports are not needed anymore?

it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
```diff
@@ -150,7 +158,57 @@ public void testTriggerInstallInNewDatabase() {
...
+            );
+
+            // drop database
+            sysSession.writeTransaction(tx -> tx.run(String.format("DROP DATABASE %s WAIT;", INIT_DB)));
+
+            // check that the trigger has been removed
+            testCallEmpty(sysSession, "CALL apoc.trigger.show($dbName)",
+                    Map.of("dbName", INIT_DB)
+            );
+        }
```
ncordon, line 210: Can you split this into another test please?

vga91: Separated

ncordon (approved):

ncordon:
Feel free to merge, we can fix the Snyk issues in another PR, as they are related to dependencies that they've just deemed are insecure

PR #349, opened 2023-03-14
[3QAbWQds] apoc.export.arrow.all ignores export file config

vga91:
Added check for procedures using file mode

### DIFF
--- a/core/src/main/java/apoc/export/arrow/ExportArrowService.java
+++ b/core/src/main/java/apoc/export/arrow/ExportArrowService.java
@@ -11,8 +11,14 @@ import org.neo4j.procedure.TerminationGuard;
 
 import java.util.stream.Stream;
 
+import static apoc.ApocConfig.APOC_EXPORT_FILE_ENABLED;
+import static apoc.ApocConfig.EXPORT_NOT_ENABLED_ERROR;
+import static apoc.ApocConfig.apocConfig;
+
 public class ExportArrowService {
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/export/arrow/ExportArrowService.java
```diff
@@ -34,6 +40,11 @@ public Stream<ByteArrayResult> stream(Object data, ArrowConfig config) {
     }
 
     public Stream<ProgressInfo> file(String fileName, Object data, ArrowConfig config) {
+        // we cannot use apocConfig().checkWriteAllowed(..) because the error is confusing
+        //  since it says "... use the `{stream:true}` config", but with arrow procedures the streaming mode is implemented via different procedures
+        if (!apocConfig().getBoolean(APOC_EXPORT_FILE_ENABLED)) {
+            throw new RuntimeException(EXPORT_TO_FILE_ARROW_ERROR);
+        }
```
gem-neo4j, line 47: Do we not have restrictions on where the file can be written to as well?

vga91: Indeed the procedure is not so tested about these cases.

I added an `ExportArrowSecurityTest` file
similar to `ExportCoreSecurityTest`, to test all these cases.

I preferred to create a separate file, instead of adding the procedures in `ExportCoreSecurityTest`,
because it works too differently (parameter order, `{stream:true}` not present, etc..)

PR #356, opened 2023-03-21
[1gsD8V8g] Add a more useful error message for export functions

gem-neo4j:
When trying to use SHOW CONSTRAINTS or SHOW INDEXES, it looks like the current user doesn't have permission to use it instead of the procedure, this adds a more generic error message.

### DIFF
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -109,6 +109,9 @@ public class Util {
 
     public static final int REDIRECT_LIMIT = 10;
 
+    public static String INVALID_QUERY_MODE_ERROR = "This procedure allows for READ-only, non-schema based queries. " +
+            "It is therefore not possible to perform writes or query the database with commands such as SHOW CONSTRAINTS/INDEXES.";
+
     public static String labelString(List<String> labelNames) {
         return labelNames.stream().map(Util::quote).collect(Collectors.joining(":"));
     }
--- a/common/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
+++ b/common/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
@@ -6,10 +6,13 @@ import org.neo4j.graphdb.*;
 import org.neo4j.graphdb.schema.ConstraintDefinition;
 import org.neo4j.graphdb.schema.IndexDefinition;
 import org.neo4j.graphdb.schema.IndexType;
+import org.neo4j.graphdb.security.AuthorizationViolationException;
 
 import java.util.*;
 import java.util.stream.Collectors;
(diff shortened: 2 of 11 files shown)


common/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
```diff
@@ -44,12 +47,16 @@ public void add(Relationship rel) {
...
-        });
+        try {
+            result.forEachRemaining(row -> {
+                for (String column : columns) {
+                    final Object value = row.get(column);
+                    graph.addToGraph(value);
+                }
+            });
+        } catch (AuthorizationViolationException e) {
+            throw new RuntimeException(INVALID_QUERY_MODE_ERROR);
```
Lojjs, line 58: Besides `apoc.export.cypher.query` and some procedures which does not take queries,  this method is also used by `apoc.export.graphml.query`. Do we want to add a test for that one as well?

common/src/main/java/apoc/util/Util.java
```diff
@@ -109,6 +109,9 @@ public class Util {
 
     public static final int REDIRECT_LIMIT = 10;
 
+    public static String INVALID_QUERY_MODE_ERROR = "This procedure allows for READ-only, non-schema based queries. " +
+            "It is therefore not possible to perform writes or query the database with commands such as SHOW CONSTRAINTS/INDEXES.";
```
Lojjs, line 113: Might be worth adding a non-schema write query to each test as well, to make sure that the write part of the error message holds.

Lojjs (changes requested):
I think the solution seems fine but there are some additions to testing I would like to see

core/src/test/java/apoc/load/LoadCoreSecurityTest.java
```diff
@@ -144,11 +144,4 @@ public void testReadSensitiveFileWorks() {
             }
         }
     }
-
-    private static void assertError(Exception e, String errorMessage, Class<? extends Exception> exceptionType, String apocProcedure) {
```
Lojjs, line 148: I think this removal has caused some unused imports

common/src/main/java/org/neo4j/cypher/export/CypherResultSubGraph.java
gem-neo4j: Ah I missed one! nice catch :)

common/src/main/java/apoc/util/Util.java
gem-neo4j: The writes are actually caught in a different place, but I liked the more generic message which is why I wrote it like this, as it still true you can't write in any of these queries

Lojjs (approved):

PR #360, opened 2023-04-01
Test that all load and import procedures respect security

vga91:


core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
```diff
@@ -0,0 +1,263 @@
...
+    public static void setUp() throws IOException {
+        TestUtil.registerProcedure(db,
+                ImportJson.class, Xml.class, ImportCsv.class, ExportGraphML.class,
+                LoadJson.class, Xml.class);
+    }
+    
+
+    @Test
+    public void testIllegalFSAccessWithDifferentApocConfs() {
+        setFileApocConfigs(true, true, true);
```
gem-neo4j, line 131: Can a comment be added to explain which each boolean is setting so the test is easier to read as is?

core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
```diff
@@ -0,0 +1,263 @@
...
+        setFileApocConfigs(false, true, true);
+        assertIpAddressBlocked();
+
+        setFileApocConfigs(false, false, false);
+        assertIpAddressBlocked();
+    }
+
+    @Test
+    public void testImportFileDisabled() {
+        setFileApocConfigs(false, false, false);
```
gem-neo4j, line 149: Can a comment be added to explain which each boolean is setting so the test is easier to read as is?

core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
gem-neo4j: Also would help explain why not all permutations are listed here, I can't directly see why we aren't also testing:

false, false, true
true, false, true
true, true, false

core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
```diff
@@ -0,0 +1,263 @@
...
+    @ClassRule
+    public static DbmsRule db = new ImpermanentDbmsRule()
+            .withSetting(GraphDatabaseSettings.load_csv_file_url_root, TEMP_FOLDER)
+            .withSetting(GraphDatabaseInternalSettings.cypher_ip_blocklist, List.of(new IPAddressString("127.168.0.0/8")));
+
+
+    @BeforeClass
+    public static void setUp() throws IOException {
+        TestUtil.registerProcedure(db,
+                ImportJson.class, Xml.class, ImportCsv.class, ExportGraphML.class,
```
gem-neo4j, line 124: Do we not want arrow procedures here as well?

gem-neo4j: Strange that the load graphml prcocedure is in the ExportGraphML class, maybe add a comment here as I was about to comment asking why an export procedure class was loaded into an import test 😅

gem-neo4j (commented):
This looks really good :) I know the card specified import and load, but how hard would it be to do this style for all our export as well? then we would be super tested :D 

Just a few comments otherwise!

core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
vga91: Actually I don't remember why I didn't test all permutations 😅... Anyway I've now added them all.

core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
vga91: Added

core/src/test/java/apoc/export/ImportAndLoadCoreSecurityTest.java
vga91: Yes sorry, I thought I also added arrow, added. 
And added a comment as well

gem-neo4j (approved):
Great, thanks for this :D Nice work!

PR #361, opened 2023-04-03
[bZQoGhFu] apoc.refactor.mergeNodes is not working with uniqueness index

vga91:
Added labels at the end of the `mergeNodes()` method, to prevent schema conflicts


Lojjs (approved):
Really neat bug fix, nice work!

PR #362, opened 2023-04-03
[MCKik4qR] Remove Relationship Constraint Feature Flag

gem-neo4j:
Removes usage of the feature flag for relationship constraints. This flag has been switched to true already, so removing this won't break, as well as allow a safe removal of the flag in the other code base :)

See [here](https://github.com/neo4j/apoc/pull/270/files) for original PR!


Lojjs (approved):
Looks like it will be ready for when the feature flag is lifted and we can get a green build

PR #366, opened 2023-04-05
[VYidKhdO] The required constraint on import key for import.json is unique

gem-neo4j:
The constraint for importing json is unique, not null and multi node key shouldn't count either. These constraints are not testable in community edition, hence the new test file.


core/src/main/java/apoc/export/json/JsonImporter.java
```diff
@@ -151,17 +153,21 @@ private void manageRelationship(Map<String, Object> param) {
...
+            checkUniquenessConstraints(labels);
             flush();
             lastLabels = labels;
         }
     }
 
-    private void checkConstraints(List<String> labels) {
+    /**
+     * The constraint for the import name should be unique to avoid duplicated imports.
+     * Node keys of size one are effectively a uniqueness constraint
```
Lojjs, line 168: I agree that node key constraints should work as well as uniqueness constraints, since a node key is a combination of uniqueness + exists constraint. But this sentence seems a bit misleading as the relation between the two type of constraints is independent of then number of properties, i.e. a composite node key constraint is also effectively a composite uniqueness constraint.

Maybe you can remove the 'size one' from this sentence and explain in a separate sentence why the constraint cannot be composite?

it/src/test/java/apoc/it/core/ImportJsonEnterpriseFeaturesTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Exception e = Assert.assertThrows(Exception.class,
+                () ->  testResult(session, "CALL apoc.import.json($file, {})", map("file", filename), (result) -> {})
+        );
+
+        String expectedMsg = format(MISSING_CONSTRAINT_ERROR_MSG, "User", "neo4jImportId");
+        TestCase.assertTrue(e.getMessage().contains(expectedMsg));
+    }
+
+    @Test
+    public void shouldWithSingularNodeKeyAsConstraint() {
```
Lojjs, line 101: Missing a verb between 'should' and 'With' in the title. Maybe 'Succeed', 'Work' or 'Import'

Lojjs (commented):
Over all it looks good, just some nitpicking on text.

core/src/main/java/apoc/export/json/JsonImporter.java
gem-neo4j: Oh but then there is still a bug left 🙃 I didn't realise we had composite uniqueness constraints 🙈 Which makes a lot of sense that we do.

I assume a constraint such as:
CREATE CONSTRAINT FOR (n:Person) REQUIRE (n.neo4jImportId, n.name) IS UNIQUE;

Allows for:
(node1: {neo4jImportId : 1, name: Fred})
(node1: {neo4jImportId : 1, name: George})

Because it is the combination of the two that is unique. But this procedure requires that *neo4jImportId* is unique! Which it isn't, will add the property key size check to that one as well if you agree with my reasoning :) (will also fix the comment to talk about composite ofc!)

gem-neo4j: Yes, will fix this up :) And actually add a test for the composite case for the uniqueness only constraints I think

Lojjs (approved):

PR #373, opened 2023-04-14
[EaALzior] runTimeboxed doesn't always work

vga91:
If the inner query throws an error, 
the `apoc.cypher.runTimeboxed(...)` continues the execution until the `timeout` terminates.


gem-neo4j (approved):
LGTM :)

PR #374, opened 2023-04-14
[ivaiMqVb] CALL apoc.export.cypher.schema(null, { stream: true }) returns nothing

vga91:
Forced `stream` and `streamStatements` to `"false"`


Lojjs (approved):
Looks good to me :)

PR #380, opened 2023-04-18
[Q8q7GXdk] apoc.export.graphml imports unwanted nodes

vga91:
The bug regard the `apoc.export.cypher` as well, since the same method it's called.

- Added `addNodes` parameter with value `false` with export procedures (i.e. not export extra-nodes) 

- Added missing tests with `nodesOfRelationships`

### DIFF
--- a/core/src/test/java/apoc/export/cypher/ExportCypherTest.java
+++ b/core/src/test/java/apoc/export/cypher/ExportCypherTest.java
@@ -1,8 +1,6 @@
 package apoc.export.cypher;
 
 import apoc.export.util.ExportConfig;
-import apoc.graph.Graphs;
-import apoc.schema.Schemas;
 import apoc.util.BinaryTestUtil;
 import apoc.util.CompressionAlgo;
 import apoc.util.MapUtil;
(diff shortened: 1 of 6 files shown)


core/src/test/java/apoc/export/cypher/ExportCypherTest.java
```diff
@@ -163,6 +159,91 @@ public void testExportQueryCypherForNeo4j() {
...
+
+        // check that `nodesOfRelationships: true` doesn't change the result
+        config.put("nodesOfRelationships", true);
+
+        commonDataAndQueryAssertions(fileName, exportQuery, params,
+                EXPECTED_REL_ONLY, 0L, 1L, 1L);
+
+        commonDataAndQueryAssertions(fileName, exportData, params,
+                EXPECTED_REL_ONLY, 0L, 1L, 1L);
+    }
```
ncordon, line 190: I'd suggest splitting this in another test. I'd advise to keep an assertion (or a group of assertions related to the same input) per tests. It's never good when a test fails and you have to dig out which input it failed for.

core/src/test/java/apoc/export/cypher/ExportCypherTest.java
```diff
@@ -163,6 +159,91 @@ public void testExportQueryCypherForNeo4j() {
...
+                "COMMIT%n");
+        String expectedWithEndNode = expectedNodesWithEndNode + EXPECTED_SCHEMA_ONLY_START + EXPECTED_REL_ONLY + EXPECTED_CLEAN_UP;
+
+        commonDataAndQueryAssertions(fileName, exportQuery, params,
+                expectedWithEndNode, 2L, 1L, 5L);
+
+        // apoc.export.cypher.data doesn't accept `nodesOfRelationships` config,
+        // so it returns the same result as the above one
+        commonDataAndQueryAssertions(fileName, exportData, params,
+                expectedWithoutEndNode, 1L, 1L, 3L);
```
ncordon, line 230: Same here please

core/src/test/java/apoc/export/cypher/ExportCypherTest.java
vga91: Changed

ncordon (approved):

PR #381, opened 2023-04-19
[ZkSGN6PO] apoc.export.graphml.all doesn't accept absolute Windows paths

vga91:
- Changed export procedures to handle path consistently with imports

- Added a note for windows user is case of a `c:/path/to/file`.

---

Tested directly on windows.
It would be great to add tests for both windows and unix path, 
but I tried adding tests to emulate a windows path (e.g. with [this one](https://github.com/google/jimfs)), 
and they don't work properly.
Maybe we can create a separate card to try to add them?

### DIFF
--- a/common/src/main/java/apoc/util/FileUtils.java
+++ b/common/src/main/java/apoc/util/FileUtils.java
@@ -38,6 +38,17 @@ import static apoc.util.Util.readHttpInputStream;
  */
 public class FileUtils {
 
+    public static String getFileUrl(String fileName) throws MalformedURLException {
+        try {
+            return new URL(fileName).getFile();
+        } catch (MalformedURLException e) {
+            if (e.getMessage().contains("no protocol")) {
+                return fileName;
+            }
+            throw e;
+        }
+    }
+
     public static StreamConnection getStreamConnection(SupportedProtocols protocol, String urlAddress, Map<String, Object> headers, String payload) throws IOException {
         switch (protocol) {
         case s3:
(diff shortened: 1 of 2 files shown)


common/src/main/java/apoc/util/FileUtils.java
```diff
@@ -97,6 +101,13 @@ public static SupportedProtocols from(String source) {
                     // otherwise we return unknown protocol: yyyyy
                     return SupportedProtocols.valueOf(new URI(source).getScheme());
                 } catch (Exception ignored) {}
+
+                // in case a Windows user write an url like `C:/User/...`
+                if (e.getMessage().contains("unknown protocol") && Util.isWindows()) {
+                    throw new RuntimeException(e.getMessage() +
+                            "\n Please note that in case of a Windows absolute path, it have to be explicit by prepending file: or without drive, " +
+                            "\n e.g. `file:C:/my/path/file` or `/my/path/file`, instead of `C:/my/path/file`");
```
gem-neo4j, line 109:
```suggestion
                             "\n Please note that for Windows absolute paths they have to be explicit by prepending 'file': or supplied without the drive, " +
                            "\n e.g. `file:C:/my/path/file` or `/my/path/file`, instead of `C:/my/path/file`");
```

Have you managed to test that this works with different drives? (I don't have a Windows machine so can't check it)

vga91: Yep, it works properly.
I changed the wording

gem-neo4j (approved):
LGTM, can you add a card for exploring how to add Windows tests like you suggested, sounds like a good idea to me :)

vga91:
> LGTM, can you add a card for exploring how to add Windows tests like you suggested, sounds like a good idea to me :)

Added trello card with id `o1mFBgUF` :)

PR #383, opened 2023-04-20
[nKMspF5h] Change developer tag in pom.xml

vga91:
Changed consistently with the neo4j one: https://github.com/neo4j/neo4j/blob/5.6/pom.xml#L33


Lojjs (approved):
Thanks for fixing!

PR #388, opened 2023-04-27
[gULAAkri] apoc.periodic.repeat doesn't always work

### LINKED ISSUE #17: The apoc.periodic.repeat and apoc.periodic.sumbit procedures don't work with some statements which return something
<a href="https://github.com/vga91"></img></a> **Issue by [vga91](https://github.com/vga91)**
_Thursday May 26, 2022 at 14:07 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/2942_

----

See also: https://community.neo4j.com/t/apoc-periodic-repeat-with-node-similarity/55782/2

Create 2 similar jobs with the `apoc.log.info` procedure. The 1st return void:
```
CALL apoc.periodic.repeat('jobOne', "CALL apoc.log.info('testOne')", 2)
```
and the 2nd return `1`:
```
CALL apoc.periodic.repeat('jobTwo', "CALL apoc.log.info('testTwo') RETURN 1", 2)
```

Note that only `jobOne` create `log.info`s

-----

The same bug regarding the `apoc.periodic.submit`, for example:
```
CALL apoc.periodic.submit('submit-2', 'CALL apoc.log.info('testThree') RETURN 1', {})
```

---

### Versions

  - OS: MacOs
  - Neo4j: 4.4.5
  - Neo4j-Apoc: 4.4.0.5

vga91:
Fixes #17

Most likely the bug occurs when there is a `RETURN ...` preceded by a void procedure with mode = "READ" (like the related gh issue)

- Added `Result::resultAsString` in order to consume results with void procedures.
- Check that the `db.prepareForReplanning` (which is a void read procedure) logs correctly

### DIFF
--- a/core/src/test/java/apoc/periodic/PeriodicTest.java
+++ b/core/src/test/java/apoc/periodic/PeriodicTest.java
@@ -40,6 +40,7 @@ import org.neo4j.graphdb.schema.Schema;
 import org.neo4j.kernel.api.KernelTransactionHandle;
 import org.neo4j.kernel.impl.api.KernelTransactions;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.logging.AssertableLogProvider;
 import org.neo4j.test.rule.DbmsRule;
 import org.neo4j.test.rule.ImpermanentDbmsRule;
 
@@ -48,6 +49,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


ncordon (approved):

core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -51,21 +53,64 @@
...
+        logProvider.clear();
+
+        // execute a periodic procedure with `CALL db.prepareForReplanning` as an inner procedure
+        db.executeTransactionally(query);
+
+        // print logs in logProvider
+        // and check that the previous `prepareForReplanning` logs a "Manual trigger for sampling all indexes"
+        assertEventually(() -> {
+            logProvider.print(System.out);
+            return logProvider.serialize().contains("Manual trigger for sampling all indexes");
```
ncordon, line 109: We are relying on something we write to the logs here in a procedure we don't control. We just need a void procedure with some effect right? 

I feel like it would be a better approach if we do a `dbms.setConfigValue` for example, fetching the value of that config afterwards.

ncordon (approved):
Good job

PR #391, opened 2023-05-03
[D2VH2D8O] Only create trigger listeners on system db.

Lojjs:
Adding listeners to each database can cause nodes from user databases to be removed, as well as cause issues with concurrent drop database commands.


ncordon (approved):

ncordon:
Good catch 🚀 . I think we should disregard the failing tests as there seems to be an API that is quite slow that's also failing in 4.4

PR #398, opened 2023-05-15
[P7cw3Wcj] Fix zip bomb attack

vga91:
It should not be possible to perform zip bomb attacks, see test `testLoadJsonShouldPreventZipBombAttack`.

But in theory, it is possible to create a compression bomb using gzip, or other formats.
I created a LimitedSizeInputStream, to limit the decompression rate of the files to 100.
But I think, instead of hardcoding to 100, the best solution is to create an apocConfig (eg `Apoc.decompression.limit`) with which you can customize the value.

---

To be added after evaluating if the pr is valid (perhaps in a new card):
- test with S3 and Hdfs
- test with arrow


common/src/main/java/apoc/ApocConfig.java
```diff
@@ -73,6 +73,7 @@ public class ApocConfig extends LifecycleAdapter {
     public static final String APOC_CONFIG_JOBS_POOL_NUM_THREADS = "apoc.jobs.pool.num_threads";
     public static final String APOC_CONFIG_JOBS_QUEUE_SIZE = "apoc.jobs.queue.size";
     public static final String APOC_CONFIG_INITIALIZER = "apoc.initializer";
+    public static final String APOC_DECOMPRESSION_LIMIT = "apoc.decompression.limit";
```
Lojjs, line 76: I would like input from the Neo4j security team on whether having this configurable or hardcode it to 100 is better

Lojjs (commented):
Looks OK to me, but I would like a second opinion on the hard-coded 100 as well as on the level in the code the checks are placed on from someone in Neo4j's security team. I will ask them to check

Lojjs (approved):
I'm happy with this now

PR #415, opened 2023-05-19
[rGkwaqu1] apoc.path.expand doesn't handle special characters in label name

vga91:
Unlike what is written in the card, 
the issue regards the `:` character, which handle compound labels.

---

In this pr I updated the `:` char so that an escape character `\` can be inserted.
In case it's ok, the documentation should be updated.

Otherwise, we could just explain better that the `:` char 
is used to handle multiple labels and that it is not possible to filter labels containing that character.

### DIFF
--- a/core/src/test/java/apoc/path/ExpandPathTest.java
+++ b/core/src/test/java/apoc/path/ExpandPathTest.java
@@ -30,16 +30,20 @@ import org.junit.ClassRule;
 import org.junit.Test;
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.Path;
+import org.neo4j.graphdb.Result;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.test.rule.DbmsRule;
 import org.neo4j.test.rule.ImpermanentDbmsRule;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import static apoc.util.Util.labelStrings;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 public class ExpandPathTest {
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/path/ExpandPathTest.java
```diff
@@ -525,4 +525,71 @@ public void testRelationshipFilterWorksWithoutTypeIncoming() {
...
+
+	@Test
+	public void testCompoundLabelWithSpecialChar() {
+		db.executeTransactionally(
+				"CREATE (n:`http://compound.example/abc#Object` {one: 'alpha'})-[:REL]->(o:`/http://www.w3.org/2002/07/owl#Class`:OwlClass:`Go:ku` {two: 'beta'})"
+		);
+
+		String pathFilter = ">|<";
+		// match using multiple labels (`http://www.w3.org/2002/07/owl#Class`, `OwlClass` and `Go:ku`)
+		String labelFilter = "//http\\://www.w3.org/2002/07/owl#Class:OwlClass:Go\\:ku|foo:bar|another";
```
Lojjs, line 566: It would also be good with a test where you e.g. have a node with label \`foo:bar\` and make sure that it is not matched by a label filter foo:bar without the escaping \\\

core/src/test/java/apoc/path/ExpandPathTest.java
```diff
@@ -525,4 +525,71 @@ public void testRelationshipFilterWorksWithoutTypeIncoming() {
 					assertEquals(1, maps.size());
 				});
 	}
+
+	@Test
+	public void testLabelWithSpecialChar() {
+		db.executeTransactionally(
+				"CREATE (n:`http://example.com/abc#Object` {one: 'alpha'})-[:REL]->(o:`http://www.w3.org/2002/07/owl#Class`:OwlClass {two: 'beta'})"
```
Lojjs, line 532: I would write the test setup slightly different so there are some nodes with e.g. the labels 
`http` +` //example.com/abc#Object`, only `http` or other similar cases that should be filtered out by the label filter. That way it would be easier to notice if there are any bugs around this solution.

Lojjs (changes requested):
Sorry for the late review. The solution looks OK, but I think it would be good with better test coverage

core/src/test/java/apoc/path/ExpandPathTest.java
vga91: Added

Lojjs (approved):
Thanks for fixing

vga91:
CI error fixed [here](https://github.com/neo4j/apoc/pull/440)

PR #422, opened 2023-05-23
[dITjfMpu] apoc.coll.indexOf unexpectedly treats collections differently than the same hardcoded list

vga91:
Potentially this behaviour can affect a lot of procedures/functions, both `apoc.coll` and elsewhere.
I changed some of them, just to discuss about a possible solution that can be good for each procedure/fun that may have the problem.
For the other cases I created a follow-up card (id: `v8imq7eo`)

I think the possible solution are:
- Util.toAnyValues(..), as done in `apoc.coll.toSet`, which transform every object in `AnyValue`
  - pros: less code to change, just wrap the lists with that method
  - cons: tranform Object in AnyValue, so e.g an hardcoded `[1,2,3]` is converted from ArrayList<Long> to long[], so it could be considered a breaking change
  - 
- Create Util method that leverage `ValueUtils.of()`, as done in `apoc.coll.indexOf` and `apoc.coll.containsAll` 
  - pros: no need to loop through lists to transform to AnyValue
  - cons: e.g with an `apoc.coll.toSet(<NodePropValue>, ... <HardcodedValue>)`, if the two `..Value` are equal, the function will return the nodeProp or the hardcoded based on which one finds first.

### DIFF
--- a/core/src/test/java/apoc/coll/CollTest.java
+++ b/core/src/test/java/apoc/coll/CollTest.java
@@ -20,6 +20,7 @@ package apoc.coll;
 
 import apoc.convert.Json;
 import apoc.util.TestUtil;
+import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
@@ -40,12 +41,19 @@ import static junit.framework.TestCase.assertTrue;
 import static org.junit.Assert.*;
 
 public class CollTest {
+    // query that procedures a list,
+    // with both entity types, via collect(..), and hardcoded one
+    private static final String QUERY_WITH_MIXED_TYPES = "MATCH (n:Test) " +
+            "WITH n ORDER BY n.a " +
+            "WITH COLLECT({something: n.something}) + { something: [] } + {something: 'alpha'} + {something: [1,2,3]} AS collection \n";
+
     private static final String QUERY_WITH_ARRAY = "CREATE (:Test {a: 1, something: 'alpha' }), " +
             "(:Test { a: 2, something: [] }), " +
-            "(:Test { a: 3, something: 'beta' })";
-    private static final String sss = "MATCH (n:Test) " +
-            "WITH n ORDER BY n.a " +
-            "WITH COLLECT({something: n.something}) as collection \n";
+            "(:Test { a: 3, something: 'beta' })," +
+            "(:Test { a: 4, something: [1,2,3] })";
+
+    public static final Map<String, String> MAP_WITH_ALPHA = Map.of("something", "alpha");
+    public static final Map<String, String> MAP_WITH_BETA = Map.of("something", "beta");
 
     @ClassRule
     public static DbmsRule db = new ImpermanentDbmsRule();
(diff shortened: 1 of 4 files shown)


core/src/test/java/apoc/coll/CollTest.java
gem-neo4j, line 1: I also did a test on another apoc.coll proc:

```
WITH COLLECT { 
    MATCH (n:Test) 
    RETURN n.something
    ORDER BY n.a}  + [[]] + [[1,2,3]] AS collection
CALL apoc.coll.elements(collection) 
YIELD _1,  _1l,
      _2,  _2l,
      _3, _3l,
      _4,  _4l,
      _5,  _5l,
      _6, _6l
RETURN _1,  _1l,
       _2,   _2l,
       _3,   _3l,
      _4,  _4l,
      _5,  _5l,
      _6, _6l
```

And it also doesn't work (lists are returned as null if they are from the collect).

There are a lot of coll funcs and procs, but I guess it would make sense for all to be fixed for this issue?

vga91: I added the `apoc.coll.elements` array handling.
I think it's only a problem of that procedure, at least regarding the `apoc.coll` ones,
since the method I changed is only called by that function.
But potentially other pieces of code could have that problem, 
so I'd create another follow-up card to investigate about that. Wdyt?

vga91: Or maybe, we can fix the `apoc.coll.elements` directly in the new follow-up card, 
together with the research/fix about other analogous problems.

core/src/test/java/apoc/coll/CollTest.java
```diff
@@ -40,6 +41,19 @@
 import static org.junit.Assert.*;
 
 public class CollTest {
+    // query that procedures a list,
+    // with both entity types, via collect(..), and hardcoded one
+    private static final String QUERY_WITH_MIXED_TYPES = "MATCH (n:Test) " +
+            "WITH n ORDER BY n.a " +
```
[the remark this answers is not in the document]
vga91: changed

core/src/test/java/apoc/coll/CollTest.java
gem-neo4j: Yep, a new card is fine :)

core/src/test/java/apoc/coll/CollTest.java
```diff
@@ -40,6 +41,22 @@
 import static org.junit.Assert.*;
 
 public class CollTest {
+    // query that procedures a list,
+    // with both entity types, via collect(..), and hardcoded one
+    private static final String QUERY_WITH_MIXED_TYPES = """
+            WITH COLLECT {
+            MATCH (n:Test)
+            RETURN {something: n.something}
+            ORDER BY n.a} + { something: [] } + {something: 'alpha'} + {something: [1,2,3]} AS collection\s
```
gem-neo4j, line 50: What is the \s here for?

gem-neo4j (approved):
LGTM :)

core/src/test/java/apoc/coll/CollTest.java
vga91: Created the follow-up card (id: `gtjPFvMD`)

core/src/test/java/apoc/coll/CollTest.java
vga91: It was just a whitespace which intellij had automatically added. 
Removed :)

PR #424, opened 2023-05-29
[fdGuLCUX] Add type constraint support

gem-neo4j:
Add support for new type constraints as well as tests.

I also noticed that the unique column was not correct for constraints that were being dropped, so fixed that (so the behaviour is the same for created/dropped constraints of the same type)

### DIFF
--- a/common/src/test/java/apoc/util/UtilTest.java
+++ b/common/src/test/java/apoc/util/UtilTest.java
@@ -1,0 +1,57 @@
+package apoc.util;
+
+import org.junit.Test;
+import org.neo4j.graphdb.schema.ConstraintType;
+import org.neo4j.graphdb.schema.IndexType;
+
+import java.util.Arrays;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+import static org.neo4j.graphdb.schema.ConstraintType.NODE_KEY;
+import static org.neo4j.graphdb.schema.ConstraintType.NODE_PROPERTY_EXISTENCE;
+import static org.neo4j.graphdb.schema.ConstraintType.NODE_PROPERTY_TYPE;
+import static org.neo4j.graphdb.schema.ConstraintType.RELATIONSHIP_KEY;
+import static org.neo4j.graphdb.schema.ConstraintType.RELATIONSHIP_PROPERTY_EXISTENCE;
+import static org.neo4j.graphdb.schema.ConstraintType.RELATIONSHIP_PROPERTY_TYPE;
+import static org.neo4j.graphdb.schema.ConstraintType.RELATIONSHIP_UNIQUENESS;
+import static org.neo4j.graphdb.schema.ConstraintType.UNIQUENESS;
+import static org.neo4j.graphdb.schema.IndexType.FULLTEXT;
+import static org.neo4j.graphdb.schema.IndexType.LOOKUP;
+import static org.neo4j.graphdb.schema.IndexType.POINT;
+import static org.neo4j.graphdb.schema.IndexType.RANGE;
+import static org.neo4j.graphdb.schema.IndexType.TEXT;
+
+public class UtilTest {
+
+    /**
+     * If any new constraints or indexes are added, this test will fail.
+     * Add the new constraints/indexes to the tests as well and update
+     * the apoc.schema.* procedures to work with them.
+     */
+    @Test
+    public void testAPOCisAwareOfAllConstraints() {
+        assertEquals(Arrays.stream(ConstraintType.values()).collect(Collectors.toSet()), Set.of(
+                UNIQUENESS,
+                NODE_PROPERTY_EXISTENCE,
+                RELATIONSHIP_PROPERTY_EXISTENCE,
+                NODE_KEY,
+                RELATIONSHIP_KEY,
+                RELATIONSHIP_UNIQUENESS,
+                RELATIONSHIP_PROPERTY_TYPE,
+                NODE_PROPERTY_TYPE
+        ));
+    }
+
+    @Test
+    public void testAPOCisAwareOfAllIndexes() {
+        assertEquals(Arrays.stream(IndexType.values()).collect(Collectors.toSet()), Set.of(
+                FULLTEXT,
+                LOOKUP,
+                TEXT,
+                RANGE,
+                POINT
+        ));
+    }
+}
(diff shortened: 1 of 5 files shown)


common/src/test/java/apoc/util/UtilTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    public void testAPOCisAwareOfAllIndexes() {
+        assertEquals(Arrays.stream(IndexType.values()).collect(Collectors.toSet()), Set.of(
+                FULLTEXT,
+                LOOKUP,
+                TEXT,
+                RANGE,
+                POINT
+        ));
+    }
```
ncordon, line 56: 👍

test-utils/src/main/java/apoc/util/TestContainerUtil.java
```diff
@@ -149,6 +149,7 @@ private static Neo4jContainerExtension createNeo4jContainer(List<ApocPackage> ap
                 .withNeo4jConfig("dbms.logs.http.enabled", "true")
                 .withNeo4jConfig("dbms.logs.debug.level", "DEBUG")
                 .withNeo4jConfig("dbms.routing.driver.logging.level", "DEBUG")
+                .withNeo4jConfig("internal.dbms.type_constraints", "true")
```
ncordon, line 152: Why do we have integration tests for this instead of unit tests?

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
```diff
@@ -164,6 +164,93 @@ private static void schemaNodeKeyAssertions(Map<String, Object> r) {
...
+            tx.run("CREATE CONSTRAINT node_prop_type_movie FOR (m:Movie) REQUIRE (m.first) IS :: INTEGER");
+            tx.commit();
+            return null;
+        });
+
+        testResult(session, CALL_SCHEMA_NODES_ORDERED, (result) -> {
+            Map<String, Object> r = result.next();
+            assertEquals("Movie", r.get("label"));
+            assertEquals(List.of("first"), r.get("properties"));
+            assertEquals(":Movie(first)", r.get("name"));
```
ncordon, line 179: This seems incorrect as name of the type constraint?

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
```diff
@@ -164,6 +164,93 @@ private static void schemaNodeKeyAssertions(Map<String, Object> r) {
...
+                assertEquals(1, expectedResult.stream().filter(
+                        c -> c.keys.containsAll(con.keys) &&
+                                c.action.equals(con.action) &&
+                                c.label.equals(con.label) &&
+                                c.unique == con.unique
+                ).toList().size());
+            }
+
+            assertEquals(expectedResult.size(), actualResult.size());
+        });
```
ncordon, line 228: Could you explain me what we are trying to test here 🙏 ?

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
```diff
@@ -674,6 +808,34 @@ public void testRelationshipKeyConstraint() {
...
+            tx.run("CREATE CONSTRAINT rel_con FOR ()-[knows:KNOWS]-() REQUIRE knows.day IS :: INTEGER");
+            tx.commit();
+            return null;
+        });
+
+        testResult(session, "CALL apoc.schema.relationships({})",
+                result -> {
+                    Map<String, Object> r = result.next();
+
+                    assertEquals("CONSTRAINT FOR ()-[knows:KNOWS]-() REQUIRE knows.day IS :: INTEGER", r.get("name"));
```
ncordon, line 823: Shouldn't this be the name of the constraint:`rel_con`?

test-utils/src/main/java/apoc/util/TestContainerUtil.java
gem-neo4j: Well they are enterprise edition only features, so thought they needed to be done this way due to that? I got errors at least when trying with unit tests, and didn't try much more than that. Is there a way to do them that you know of?

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
```diff
@@ -164,6 +164,93 @@ private static void schemaNodeKeyAssertions(Map<String, Object> r) {
...
+            assertEquals("Movie", r.get("label"));
+            assertEquals(List.of("first"), r.get("properties"));
+            assertEquals(":Movie(first)", r.get("name"));
+            assertEquals("", r.get("status"));
+            assertEquals("NODE_PROPERTY_TYPE", r.get("type"));
+            final String expectedUserDescConstraint = "name='node_prop_type_movie', type='NODE PROPERTY TYPE', schema=(:Movie {first}), propertyType=INTEGER";
+            Assertions.assertThat(r.get("userDescription").toString()).contains(expectedUserDescConstraint);
+
+            assertFalse(result.hasNext());
+        });
```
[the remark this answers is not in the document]
gem-neo4j: There is actually a @before on all that remove all constraints, so I'll clean up the tests so as not to do extra work :)

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
gem-neo4j: I gave up questioning APOC :P It is consistent with how it works today (see docs here[https://neo4j.com/docs/apoc/current/overview/apoc.schema/apoc.schema.nodes/](url)). I agree it is weird, but I wouldn't change it here, would you like me to make a card for it?

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
gem-neo4j: So I copied the style for each type of test as I figured being consistent per test type was less confusing than per constraint type. But what happens with this one is that above the loop is written the 2 expected results, and this checks that both results exist in the result set. The first constraint is in the constraint list of the CALL so should be kept, the second constraint is not mentioned so should be dropped. So the test checks one is kept and one is dropped :)

test-utils/src/main/java/apoc/util/TestContainerUtil.java
ncordon: Ahhhh, I overlooked DDL commands are enterprise only. All good 👍

it/src/test/java/apoc/it/core/SchemasEnterpriseFeaturesTest.java
ncordon: Let's make a card for it please

ncordon (approved):

PR #425, opened 2023-06-02
[KUEoGBey] Sanitize text input and add quotes to allow special characters in Atomic procedures

gem-neo4j:
Properties can use special characters, and they should be able to be updated via Atomic.update (and all other atomic procedures)


nadja-muller (approved):
Looks good 👍

PR #428, opened 2023-06-05
[Au0RCa4z] Upgrade dependencies

ncordon:
Supersedes #427


build.gradle
```diff
@@ -132,5 +132,5 @@ apply from: "licenses-source-header.gradle"
 ext {
     publicDir =  "${project.rootDir}"
     neo4jVersionEffective = project.hasProperty("neo4jVersionOverride") ? project.getProperty("neo4jVersionOverride") : "5.9.0"
-    testContainersVersion = '1.17.6'
+    testContainersVersion = '1.18.3'
```
Lojjs, line 135: If I search for 1.17.6 I get among other hits, this line in LICENSES.txt:

testcontainers-1.17.6.jar

Do we need to do anything so the license files are updated when we update dependencies and the build fails otherwise, like in the monorepo? cc @gem-neo4j

test-utils/build.gradle
```diff
@@ -53,7 +53,7 @@ dependencies {
     implementation group: 'org.gradle', name: 'gradle-tooling-api', version: '7.3'
     implementation group: 'org.jetbrains', name: 'annotations', version: "17.0.0"
     implementation group: 'com.amazonaws', name: 'aws-java-sdk-s3', version: '1.12.425'
-    implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-csv', version: '2.13.2'
+    implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-csv', version: '2.13.5'
```
Lojjs, line 56: We could try to upgrade to 2.14.1 instead, because we already get the transitively.

Lojjs (changes requested):
I think we need to do something about the LICENSES file

build.gradle
gem-neo4j: Yes, good point. The build is meant to fail on new files being added without a license header as well (although this seems to only fail locally with ./gradlew build as I just tried it and it failed but this PR is fine 🤔). I'll look into this when I have time!

Lojjs: Ok maybe we can create a card for both those things?

gem-neo4j: https://trello.com/c/VpC8nqNz/2264-add-build-failure-on-license-file-change

test-utils/build.gradle
ncordon: That's not correct either actually. According to `./gradlew test-utils:dependencies` the one it gets transitively is `2.15.0`.

build.gradle
ncordon: I know what's failing, I can fix it in a separate pr. Thanks for spotting it 🤗

Lojjs: I think @gem-neo4j is looking into it already

test-utils/build.gradle
Lojjs: Oh I might have checked with `./gradlew core:dependencies`, but that is even better

PR #431, opened 2023-06-06
[drqYtb6r] apoc.cypher.run* procedures don't return results

vga91:
The entities have to be rebound, i.e. leveraging the internal id, like explained [in this example](https://neo4j.com/developer/kb/a-significant-change-in-apoc-periodic-iterate-in-apoc-4-0/).

- Added rebind in Cypher.java
- Moved EntityUtil from extended to common
- Added dockerized enterprise tests, since with embedded db it works correctly
- Added test-utils method, in order to be used by extended tests as well (see https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/3615)
- Extracted `importFolder` as a static attribute (used by extended tests), and changed path by adding the `baseDir/build` prefix, to make git ignore files created in extended

### DIFF
--- a/it/src/test/java/apoc/it/core/CypherEnterpriseTest.java
+++ b/it/src/test/java/apoc/it/core/CypherEnterpriseTest.java
@@ -1,19 +1,29 @@
 package apoc.it.core;
 
 import apoc.util.Neo4jContainerExtension;
+import org.assertj.core.api.Assertions;
+import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.neo4j.driver.Session;
+import org.neo4j.driver.types.Node;
 
 import java.util.List;
 import java.util.Map;
 
+import static apoc.cypher.CypherTestUtil.CREATE_RESULT_NODES;
+import static apoc.cypher.CypherTestUtil.CREATE_RETURNQUERY_NODES;
 import static apoc.cypher.CypherTestUtil.SET_AND_RETURN_QUERIES;
+import static apoc.cypher.CypherTestUtil.SET_NODE;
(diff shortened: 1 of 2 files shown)


it/src/test/java/apoc/it/core/CypherEnterpriseTest.java
```diff
@@ -0,0 +1,51 @@
...
+    public void testRunManyWithSetAndResults() {
+        String query = "CALL apoc.cypher.runMany($statement, {})";
+        Map<String, Object> params = Map.of("statement", SET_AND_RETURN_QUERIES);
+
+        testRunProcedureWithSetAndReturnResults(session, query, params);
+    }
+
+    @Test
+    public void testRunManyWithResults() {
+        String query = "CALL apoc.cypher.runMany($statement, {})";
```
gem-neo4j, line 45: Can we have tests for the other apoc.cypher.* procedures, I assume this also fails for apoc.cypher.runManyReadOnly perhaps? and maybe others e.g apoc.cypher.doIt (Not sure, I haven't tested them out)

gem-neo4j (commented):
Looks good overall, just curious about the other apoc.cypher.* procedures that might present the same issue :)

it/src/test/java/apoc/it/core/CypherEnterpriseTest.java
vga91: Good point.
I added test for other procedure, i.e. `apoc.cypher.runManyReadOnly`, `apoc.cypher. doIt`, `apoc.cypher. runWrite` and `apoc.cypher. run`.

vga91:
The failing test seems due to neo4j bump and is not related to this pr.
There is already a trello card on this

PR #433, opened 2023-06-09
[NOID] Rename apoc.atomic.* parameter times to retryAttempts

gem-neo4j:
This update comes from a request from docs as it was found to be confusing. It is not breaking as parameters cannot be referenced by name in Cypher, so is purely to make the signature more understandable and match changes to docs.


vga91 (approved):
LGTM :)

PR #435, opened 2023-06-09
[EdwwgWgR] Stream schema results in batches of batchSize

gem-neo4j:
The nodes and relationships weren't streaming in batches which caused an out of memory error as APOC tried to hold all the nodes/rels in memory.

Differences:

As we take nodes sequentially, there may be more groups of unwinds as previously all nodes of that label was found and in one unwind vs now they may end up in different statements due to batching (similar for rels)
There was a check for rels that assumed we knew all rels and could optimize on whether or not to filter on id, however, a batch only knows about that batch and not all rels, so that assumption no longer holds
Schema will stream before moving on to nodes, this was intended by the looks of it already, just missed the call to accept the batch
There was a bug around the counting of needed unique labels as the for loop didn't stop when it hit true (found unique), this has been corrected
I didn't add an out of memory test as I felt like this would just slow our testing down. But I did test this manually on a graph which was roughly 2.5 million nodes and 2 million rels (2x bigger than db which was reported for this to be failing on). The version without the fix doesn't work, the version with the fix did :)


common/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
```diff
@@ -312,95 +329,109 @@ private boolean isBatchMatch(ExportConfig exportConfig, AtomicInteger batchCount
 	}
 
 	public void buildStatementForRelationships(String relationshipClause,
-											   String setClause, Iterable<Relationship> relationship,
-											   Map<String, Set<String>> uniqueConstraints, ExportConfig exportConfig,
+											   String setClause,
+											   Iterable<Relationship> relationship,
+											   Map<String, Set<String>> uniqueConstraints,
+											   ExportConfig exportConfig,
 											   PrintWriter out, Reporter reporter,
```
Lojjs, line 336:
```suggestion
											   PrintWriter out, 
											   Reporter reporter,
```

common/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
```diff
@@ -312,95 +329,109 @@ private boolean isBatchMatch(ExportConfig exportConfig, AtomicInteger batchCount
...
+			String start = "start";
+			String end = "end";
+			AtomicInteger relCount = new AtomicInteger(0);
+			groupedData.forEach((path, relationshipList) -> {
+				AtomicInteger unwindCount = new AtomicInteger(0);
+				final int relSize = relationshipList.size();
+				relCount.addAndGet(relSize);
+				final Relationship last = relationshipList.get(relSize - 1);
+				for (int index = 0; index < relationshipList.size(); index++) {
+					Relationship rel = relationshipList.get(index);
```
Lojjs, line 389:
```suggestion
				for (Relationship rel : relationshipList) {
```

common/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
```diff
@@ -202,68 +202,85 @@ public String mergeStatementForRelationship(CypherFormat cypherFormat, Relations
...
+			};
+			Map<Map.Entry<Set<String>, Set<String>>, List<Node>> groupedData =
+					StreamSupport.stream(
+							com.google.common.collect.Iterables.limit(
+									com.google.common.collect.Iterables.skip(nodes, totalNodeCount.get()),
+									exportConfig.getBatchSize()).spliterator(),
+									true)
+					.collect(Collectors.groupingByConcurrent(keyMapper));
+
+			if (nodesInBatch.get() < exportConfig.getBatchSize()) shouldContinue = false;
```
Lojjs, line 237: I had to think quite a bit to realize that this was correct, so maybe it would be worth to add an explanatory code comment about the nodesInBatch counter. Same goes for `relsInBatch`

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -388,11 +389,10 @@ private long countArtificialUniques(Iterable<Node> n) {
     private long getArtificialUniques(Node node, long artificialUniques) {
         Iterator<Label> labels = node.getLabels().iterator();
         boolean uniqueFound = false;
-        while (labels.hasNext()) {
+        while (labels.hasNext() && !uniqueFound) {
```
Lojjs, line 392: Was this bug fix only a speed-up or can we add a test for it?

Lojjs (changes requested):
Overall, it looks fine. Just had two small concerns and some further clean-up that could be done if you like

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
gem-neo4j: It was both I guess, I added it for the speed, although the bug wouldn't have affected the correctness of the outputted queries. I can add a test for it though :)

Lojjs (approved):

PR #437, opened 2023-06-12
[NOID] Re-enable docker version override

ncordon:
This reverts #434

## Why

Because what it was making it fail should have already been fixed at TeamCity level.


gem-neo4j (approved):
Thanks!

PR #445, opened 2023-06-26
[NOID] Snyk Guava updates

gem-neo4j:
Update Guava and consequent license files

### DIFF
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -48,8 +48,8 @@ Apache-2.0
   docker-java-transport-3.3.0.jar
   docker-java-transport-zerodep-3.3.0.jar
   ehcache-3.3.1.jar
+  error_prone_annotations-2.18.0.jar
   error_prone_annotations-2.2.0.jar
-  error_prone_annotations-2.7.1.jar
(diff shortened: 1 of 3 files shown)


LICENSES.txt
```diff
@@ -58,7 +58,7 @@ Apache-2.0
   gradle-tooling-api-7.3.jar
   gson-2.9.0.jar
   guava-27.0-jre.jar
-  guava-31.0.1-jre.jar
+  guava-32.0.1-jre.jar
```
Lojjs, line 61: Why do we have double lines for all these dependencies? I s it because we have a transitive guava-27.0-jre.jar (from hadoop-common) that is overridden by guava-32?

Lojjs (approved):
I think it looks alright, just had a question

LICENSES.txt
gem-neo4j: I believe it is because the LICENSES is for both core and common, so core has 32 and common has 27 (due to Hadoop)

Lojjs: Ok, makes sense

PR #446, opened 2023-06-26
[NOID] Snyk Hadoop updates

gem-neo4j:
Note: the comment to check javax.servlet.jsp:jsp-api:2.1 was done, and it hasn't updated so the comment stays for now

### DIFF
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -41,9 +41,9 @@ Apache-2.0
   commons-math3-3.6.1.jar
   commons-net-3.9.0.jar
   commons-text-1.10.0.jar
-  curator-client-4.2.0.jar
-  curator-framework-4.2.0.jar
-  curator-recipes-4.2.0.jar
+  curator-client-5.2.0.jar
+  curator-framework-5.2.0.jar
+  curator-recipes-5.2.0.jar
   docker-java-api-3.3.0.jar
   docker-java-transport-3.3.0.jar
   docker-java-transport-zerodep-3.3.0.jar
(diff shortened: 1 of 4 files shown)


LICENSES.txt
```diff
@@ -170,28 +170,29 @@ Apache-2.0
...
+  netty-handler-ssl-ocsp-4.1.89.Final.jar
   netty-resolver-4.1.93.Final.jar
-  netty-resolver-dns-4.1.77.Final.jar
-  netty-resolver-dns-classes-macos-4.1.77.Final.jar
-  netty-resolver-dns-native-macos-4.1.77.Final-osx-aarch_64.jar
-  netty-resolver-dns-native-macos-4.1.77.Final-osx-x86_64.jar
+  netty-resolver-dns-4.1.89.Final.jar
+  netty-resolver-dns-classes-macos-4.1.89.Final.jar
+  netty-resolver-dns-native-macos-4.1.89.Final-osx-aarch_64.jar
+  netty-resolver-dns-native-macos-4.1.89.Final-osx-x86_64.jar
```
Lojjs, line 195: Probably there is a sensible reason but I cannot find these 2 dependencies when I list dependencies locally 🤷‍♀️

Lojjs (approved):
Tried to verify locally. There is 2 dependencies I cannot see but, I think it is probably ok as they have the same new version as the others

LICENSES.txt
gem-neo4j: I find: io.netty:netty-resolver-dns-native-macos:4.1.89.Final and am not sure exactly how it works with splitting out jars to -osx etc 🤔

Lojjs: Same, I am not to worried about it though so please go ahead and merge

PR #449, opened 2023-07-03
Annotate unsafe procedures with @NotThreadSafe

pontusmelke:
Annotate unsafe procedures with @NotThreadSafe

## Proposed Changes

When run with parallel runtime not all operations on injected instances of `Transaction` and `GraphDatabaseService` are supported. By annotating these with `@NotThreadSafe` we can avoid runtime failures and we detect already at planning that we cannot run these procedures with parallel runtime.


pontusmelke:
replaced by https://github.com/neo4j/apoc/pull/452

PR #450, opened 2023-07-03
[TFMonhHx] Forces new version of netty

ncordon:
## What
Bumps the netty version


core/build.gradle
```diff
@@ -49,13 +49,22 @@ dependencies {
...
+
+    def arrowExclusions = {
         exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'
         exclude group: 'com.fasterxml.jackson.core', module: 'jackson-annotations'
         exclude group: 'com.fasterxml.jackson.core', module: 'jackson-databind'
         exclude group: 'io.netty', module: 'netty-common'
     }
 
+    implementation group: 'org.apache.arrow', name: 'arrow-vector', version: '12.0.1', arrowExclusions
+    implementation group: 'org.apache.arrow', name: 'arrow-memory-netty', version: '12.0.1', arrowExclusions
```
gem-neo4j, line 61: Don't forget to update LICENSES :)

ncordon: This puzzled me cause I thought the CI was checking for those because of the work you did. I think you were right we need to add an extra step: https://github.com/neo4j/apoc/pull/451

gem-neo4j: Hmm this is weird, because I tested it out by pushing with missing parts and it had worked then 🤔

LICENSES.txt
```diff
@@ -58,64 +58,64 @@ Apache-2.0
...
-  hadoop-mapreduce-client-app-3.3.6.jar
-  hadoop-mapreduce-client-common-3.3.6.jar
-  hadoop-mapreduce-client-core-3.3.6.jar
-  hadoop-mapreduce-client-hs-3.3.6.jar
-  hadoop-mapreduce-client-jobclient-3.3.6-tests.jar
-  hadoop-mapreduce-client-jobclient-3.3.6.jar
-  hadoop-mapreduce-client-shuffle-3.3.6.jar
-  hadoop-minicluster-3.3.6.jar
-  hadoop-registry-3.3.6.jar
+  hadoop-annotations-3.3.5.jar
```
gem-neo4j, line 64: Guava and Hadoop have all dropped, are you on the latest dev?

ncordon: Arrrrrgggg. At least the license step seems to be catching it now

LICENSES.txt
```diff
@@ -185,6 +190,7 @@ Apache-2.0
   netty-codec-stomp-4.1.89.Final.jar
   netty-codec-xml-4.1.89.Final.jar
   netty-common-4.1.93.Final.jar
+  netty-common-4.1.94.Final.jar
```
ncordon, line 193: Any hint on why do we get two versions in the LICENSES @gem-neo4j even if we are overwritting with the most recent one?

gem-neo4j: The LICENSE file is a mix of core and common, so I believe common doesn't have 4.1.94 yet in its dependency tree

gem-neo4j: Might be from test-utils actually, I see arrow is on 10.0.1 there, should that be updated?

ncordon: It's not that, I've tried overriding there as well. This should heal automatically anyway when the database updates to 4.1.94

gem-neo4j: Yeah, should do :)

gem-neo4j (approved):

PR #452, opened 2023-07-04
[NOID] Annotate unsafe procedures with @NotThreadSafe

pontusmelke:
[NOID] Annotate unsafe procedures with @NotThreadSafe

## Proposed Changes

When run with parallel runtime not all operations on injected instances of `Transaction` and `GraphDatabaseService` are supported. By annotating these with `@NotThreadSafe` we can avoid runtime failures and we detect already at planning that we cannot run these procedures with parallel runtime.

NOTE: the `NotThreadSafe` annotation has been merged in dev but I guess we need to wait for it to become available?

### DIFF
--- a/core/src/main/java/apoc/coll/Coll.java
+++ b/core/src/main/java/apoc/coll/Coll.java
@@ -59,7 +59,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
-
+import org.neo4j.procedure.NotThreadSafe;
 import static apoc.util.Util.containsValueEquals;
 import static apoc.util.Util.toAnyValues;
 import static java.util.Arrays.asList;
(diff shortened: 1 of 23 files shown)


core/src/main/java/apoc/coll/Coll.java
```diff
@@ -149,6 +149,7 @@ public Double avg(@Name("coll") List<Number> list) {
         }
         return (avg/(double)list.size());
     }
+    @NotThreadSafe
```
ncordon, line 152: Are we marking this not thread safe because it tries to execute with `runtime=slotted`?

ncordon (commented):
What's the criteria to mark something as not thread safe? 

For example in `coll.min`, is it that we are trying to execute a transaction against the slotted runtime?

Is there any way we can test the changes are working correctly and that we haven't missed any annotations that are needed?

pontusmelke:
> What's the criteria to mark something as not thread safe?
> 
> For example in `coll.min`, is it that we are trying to execute a transaction against the slotted runtime?
> 
> Is there any way we can test the changes are working correctly and that we haven't missed any annotations that are needed?

Not all operations on injected `Transaction` or `GraphDatabaseService` are supported. For example you are not allowed to initiate new transactions nor execute cypher statements when being called from the parallel runtime. In order to verify I think we would need to run all tests with all available runtimes but I don't see any easy way of doing this change here.

ncordon (approved):

PR #453, opened 2023-07-05
[ZIv6F2QY] Allow Command Expansion APOC

gem-neo4j:
Depends on Neo4j PR for reading expandCommands.

Follows what Neo4j has:
* apoc.conf must have certain permissions only (https://neo4j.com/docs/operations-manual/current/configuration/neo4j-conf/#neo4j-conf-expand-commands)
* if the file contains a command but command expansion is not on, it will error
* if a command is invalid, it will error

### DIFF
--- a/common/src/main/java/apoc/ApocConfig.java
+++ b/common/src/main/java/apoc/ApocConfig.java
@@ -20,6 +20,8 @@ package apoc;
 
 import apoc.export.util.ExportConfig;
 import inet.ipaddr.IPAddressString;
+
+import java.io.File;
 import java.util.Map;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.PropertiesConfiguration;
@@ -40,6 +42,7 @@ import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 import org.neo4j.logging.Log;
 import org.neo4j.logging.NullLog;
 import org.neo4j.logging.internal.LogService;
+import org.neo4j.util.Preconditions;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
@@ -52,6 +55,7 @@ import java.util.List;
 import java.util.stream.Stream;
 
 import static apoc.util.FileUtils.isFile;
+import static java.lang.String.format;
 import static org.neo4j.configuration.BootloaderSettings.lib_directory;
 import static org.neo4j.configuration.BootloaderSettings.run_directory;
 import static org.neo4j.configuration.GraphDatabaseSettings.SYSTEM_DATABASE_NAME;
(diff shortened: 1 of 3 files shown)


common/src/main/java/apoc/ApocConfig.java
```diff
@@ -140,13 +154,34 @@ public Configuration getConfig() {
...
 
+    private String evaluateIfCommand(String settingName, String entry) {
+        if (Config.isCommand(entry)) {
+            Preconditions.checkArgument(
+                    expandCommands,
+                    format(
+                            "%s is a command, but config is not explicitly told to expand it. (Missing --expand-commands argument?)",
+                            entry));
+            String str = entry.trim();
+            String command = str.substring(2, str.length() - 1);
```
ncordon, line 165: Does neo4j only admit command expansion with `$(command)` or is there other way? I'm asking because in bash I think you can also enclose in backticks:

These two produce the same result:

```
echo hola $(echo 'adios')
```

```
echo hola `echo 'adios'`
```

common/src/test/java/apoc/ApocConfigTest.java
```diff
@@ -33,55 +41,158 @@
...
+
+    @Test
+    public void testApocConfWithInvalidExpandCommands() throws Exception {
+        setApocConfigSystemProperty();
+
+        String invalidExpandLine = "command.expansion.3=$(fakeCommand 3 + 3)";
+        addLineToApocConfig(invalidExpandLine);
+
+        RuntimeException e = assertThrows(RuntimeException.class, apocConfig::init);
+        String expectedMessage = "java.io.IOException: Cannot run program \"fakeCommand\": error=2, No such file or directory";
```
ncordon, line 141: Could we wrap this in a better exception so someone knows what config option failed to expand exactly?

gem-neo4j: So this exception would be written in the logs on startup, which causes Neo to not start, and directly before the line:

`            log.info("Executing external script to retrieve value of setting " + settingName);
`

Gets printed, so the user should see that and the exception. Is that what you were thinking? I can also catch the exception and add more info, but then it will behave differently to Neo (this error comes from Neo)

ncordon: Yep, the current log is good enough then in my opinion

ncordon (approved):
Nice work! 😄

common/src/main/java/apoc/ApocConfig.java
```diff
@@ -112,9 +117,18 @@ public class ApocConfig extends LifecycleAdapter {
...
+    private Duration commandEvaluationTimeout;
+
     public ApocConfig(Config neo4jConfig, LogService log, GlobalProcedures globalProceduresRegistry, DatabaseManagementService databaseManagementService) {
         this.neo4jConfig = neo4jConfig;
         this.blockedIpRanges = neo4jConfig.get(GraphDatabaseInternalSettings.cypher_ip_blocklist);
+        this.commandEvaluationTimeout = neo4jConfig.get(GraphDatabaseInternalSettings.config_command_evaluation_timeout);
+        if (this.commandEvaluationTimeout == null) {
+            this.commandEvaluationTimeout = GraphDatabaseInternalSettings.config_command_evaluation_timeout.defaultValue();
+        }
+        this.expandCommands = neo4jConfig.expandCommands();
```
irene221b, line 131: At this point, do we already know that config for expanding is set? Probably we shouldn't even try if the option is off.

gem-neo4j: I'm not sure I follow, expand command can only be set on startup, so it would already be set as Neo4j starts APOC with the Neo4J config 🤔 Which part shouldn't we try?

irene221b: Sorry, my misunderstanding of the code, ignore this comment.

PR #457, opened 2023-07-14
[NOID] Remove overrideImplementation argument

mnssn:
.. that will be removed from Neo4j. This has hidden two errors:

* BigGraphTest registered the ImportCSV methods twice.
* TriggeredNewProcedures registered the Nodes methods twice, but also
  attempted to register things on different databases (which is not
  possible at the time of writing).


gem-neo4j (approved):
LGTM :)

PR #460, opened 2023-07-24
[7BVOtcmX] Add debug statement to see if we can figure out cause of flaky test

gem-neo4j:
The test is failing due to missing results, it might be that the actual procedure is broken instead, but not reproducible locally. I have added some debug statements to see if it will help us figure out what is happening.

Feel free to suggest another idea if you don't agree with this one :)


Lojjs (approved):
Looks OK to me

PR #463, opened 2023-07-31
[Ko3jddbl] Add isDeprecated to apoc.help

gem-neo4j:
Add isDeprecated now that SHOW FUNCTIONS also supports this.


Lojjs (approved):
Nice :)

PR #465, opened 2023-08-03
[DranV6rq] Add IndexType.VECTOR to recognised index types

parnmatt:
Adding the EAP Vector Indexes into Neo4j adds a new enum value to `IndexType`.
This particular test does a runtime check that it is explicitly aware of each `IndexType`, and thus fails when reaching `VECTOR`.


Lojjs:
Replaced by https://github.com/neo4j/apoc/pull/467

PR #466, opened 2023-08-03
[NOID] Fix potential ApocConfigTest issues by using a separate conf file for command expansion

gem-neo4j:
I had seen a test failure due to the commands being seen when the apocConf was used elsewhere, so this makes sure that this test reads a special file instead.

### DIFF
--- a/common/src/test/java/apoc/ApocConfigTest.java
+++ b/common/src/test/java/apoc/ApocConfigTest.java
@@ -18,19 +18,11 @@
  */
 package apoc;
 
-import com.google.common.collect.Sets;
-import org.assertj.core.api.Assertions;
 import org.junit.Before;
 import org.junit.Test;
 
 import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.nio.file.Files;
-import java.nio.file.attribute.PosixFilePermission;
 import java.util.Collections;
-import java.util.Set;
 
 import org.neo4j.configuration.Config;
 import org.neo4j.configuration.GraphDatabaseSettings;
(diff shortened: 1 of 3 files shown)


common/src/test/java/apoc/ApocConfigTest.java
```diff
@@ -76,7 +77,8 @@ public void setup() throws Exception {
         when(neo4jConfig.expandCommands()).thenReturn(true);
 
         apocConfigFile = new File(getClass().getClassLoader().getResource("apoc.conf").toURI());
-        Files.setPosixFilePermissions(apocConfigFile.toPath(), permittedFilePermissionsForCommandExpansion);
+        apocConfigCommandExpansionFile = new File(getClass().getClassLoader().getResource("apoc-config-command-expansion/apoc.conf").toURI());
```
ncordon, line 80: Any reason why not to use the same `apoc.conf` file for this class?

gem-neo4j: Because then I don't need to add and remove the command line from the file for every other test

ncordon: It's more difficult to reason which tests use what config file. I'd rather if some tests depend on a specific file in the filesystem, then we have a file per test class. You could also extract those tests to another class `ApocConfigExpansion`?

gem-neo4j: Okay, sure :)

ncordon: Thanks 😄

ncordon (approved):

PR #467, opened 2023-08-03
[NOID] Ignore vector indexes when getting indexes from Neo4j.

Lojjs:


gem-neo4j (approved):
This looks good to me I think :)

PR #476, opened 2023-08-16
[q5Vf4pi2] Fix broken GeoCode tests

gem-neo4j:
The count of returned results had changed, hence them starting to break. I found the tests very hard to read so I rewrote them as well to be easier to understand.


Lojjs (approved):
Looks a bit clearer. Just a small comment which you do not have to fix if you not want to.

Lojjs (approved):

PR #477, opened 2023-08-17
[YGQfavsP] Remove dependency on external websites in load tests

gem-neo4j:
There are some flaky tests which fail due to github returning 502. This causes flaky tests. Instead of assuming github should always be up, I have introduced a simple HTTP server to the tests to still test http calls, but be more reliable.

I also did some warning clean up, gotta get APOC clean 🧹

### DIFF
--- a/core/src/test/java/apoc/load/SimpleHttpHandler.java
+++ b/core/src/test/java/apoc/load/SimpleHttpHandler.java
@@ -49,7 +49,6 @@ public class SimpleHttpHandler implements HttpHandler {
     }
     public void handle(HttpExchange t) throws IOException {
         URI uri = t.getRequestURI();
-        System.out.println("looking for: "+ uri.getPath());
         String path = uri.getPath().substring(1);
 
         File file;


core/src/test/java/apoc/load/SimpleHttpHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        MIME_MAP.put("json", "application/json");
+        MIME_MAP.put("xml", "application/xml");
+        MIME_MAP.put("zip", "application/zip");
+        MIME_MAP.put("tgz", "application/x-gzip");
+        MIME_MAP.put("gz", "application/x-gzip");
+        MIME_MAP.put("txt", "text/plain");
+    }
+    public void handle(HttpExchange t) throws IOException {
+        URI uri = t.getRequestURI();
+        System.out.println("looking for: "+ uri.getPath());
```
Lojjs, line 52: Unless required for debugging or so we usually not leave prints in the code. Do you want it left?

core/src/test/java/apoc/load/SimpleHttpHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        String path = uri.getPath().substring(1);
```
Lojjs, line 53: Why the .substring(1)?

core/src/test/java/apoc/load/SimpleHttpHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        if (dotIndex >= 0) {
+            return basename.substring(dotIndex + 1);
+        } else {
+            return "";
+        }
+    }
+
+    private static String lookupMime(String path) {
+        String ext = getExt(path).toLowerCase();
+        return MIME_MAP.getOrDefault(ext, "application/octet-stream");
```
Lojjs, line 100: Not really a comment but this file made me google for MIME and octet-stream. I learned something new 😁

Lojjs (approved):
Great work! I also learnt some stuff by reviewing the SimpleHttpHandler, which is always a plus :) Comments are just a println that I don't know if we want to keep and one detail I didn't understand.

core/src/test/java/apoc/load/SimpleHttpHandler.java
gem-neo4j: Hehe was my debugging, forgot to remove 😅

core/src/test/java/apoc/load/SimpleHttpHandler.java
gem-neo4j: Because the front slash will break the resource lookup, so I remove it here

PR #481, opened 2023-08-28
[uSvFQccG] Make ExportCypherTest more robust.

Lojjs:
Things like the order in UNWIND lists and the exact ids could vary, which is not a problem for the functionality but cause flakiness of the test.

I focused on the test that was marked as flaky, but this could be done for more of the export tests as a follow-up cleanup/remove risk of flakiness task if we like to.


gem-neo4j (approved):
I think this is a good way to deal with it :) Maybe make a cleanup card somewhere in APOC trello that points here so if we do get a flaky test card for another area we have links pointing to this solution?

Another thought, and I know this wasn't to do with this PR, but I find tests naming ids of github issues hard to understand, would be nice if we either had a comment explaining the original issue or a more explanatory name, nothing for you to do here, just a rant :P

PR #499, opened 2023-09-27
[Gc8EZpDO] Clean up some usages of getId() which is deprecated

gem-neo4j:
getId() is deprecated, so slowly clearing it out of APOC, here are some more usage removals as well as some general tidy up 🧹

Docs: https://github.com/neo4j/docs-apoc/pull/202


Lojjs (changes requested):
I am missing tests for elementId for apoc.algo.cover and apoc.refactor.collapseNode, otherwise it looks fine

Lojjs (approved):
Nice cleanup!

PR #503, opened 2023-10-03
[pLz5YbCW] add checks for node labels and relationship types for apoc.merge procedures

nadja-muller:
apoc.merge.nodes:
- if labelNames null is newly considered as an empty label
- if labelNames contains null values, it is considered as incorrect input
- if labelNames contains among others an empty string, it is considered as incorrect input

apoc.merge.relationships:
- if relationship type is null, it is considered as incorrect input
- if relationship type is empty, it is considered as incorrect input

### DIFF
--- a/core/src/main/java/apoc/merge/Merge.java
+++ b/core/src/main/java/apoc/merge/Merge.java
@@ -35,7 +35,6 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static apoc.util.Util.labelString;
 import static java.util.Collections.emptyMap;
 
 public class Merge {
@@ -83,16 +82,26 @@ public class Merge {
     }
 
     private Result getNodeResult(List<String> labelNames, Map<String, Object> identProps, Map<String, Object> props, Map<String, Object> onMatchProps) {
-        if (identProps ==null || identProps.isEmpty()) {
+        if (identProps == null || identProps.isEmpty()) {
             throw new IllegalArgumentException("you need to supply at least one identifying property for a merge");
         }
 
-        String labels = labelString(labelNames);
+        if (labelNames != null &&
+                (labelNames.contains(null) || (labelNames.size() == 1 && labelNames.get(0).isEmpty()))) {
+            throw new IllegalArgumentException("The list of label names contained a null value. If you wish to merge a node without a label, pass an empty list instead.");
+        }
+
+        // labelNames = [""]
+        if (labelNames != null && labelNames.size() == 1 && labelNames.get(0).isEmpty()) {
+            throw new IllegalArgumentException("The list of label cannot be empty. If you wish to merge a node without a label, pass an empty list instead.");
+        }
+
+        String labels = labelNames != null ? ":" + labelNames.stream().map(Util::quote).collect(Collectors.joining(":")) : "";
 
         Map<String, Object> params = Util.map("identProps", identProps, "onCreateProps", props, "onMatchProps", onMatchProps);
         String identPropsString = buildIdentPropsString(identProps);
 
-        final String cypher = "MERGE (n:" + labels + "{" + identPropsString + "}) ON CREATE SET n += $onCreateProps ON MATCH SET n += $onMatchProps RETURN n";
+        final String cypher = "MERGE (n" + labels + "{" + identPropsString + "}) ON CREATE SET n += $onCreateProps ON MATCH SET n += $onMatchProps RETURN n";
         return tx.execute(cypher, params);
     }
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/merge/MergeTest.java
```diff
@@ -136,23 +124,59 @@ public void testMergeWithEmptyIdentityPropertiesShouldFail() {
             }
         }
     }
-    
+
+    @Test
+    public void testMergeNodeWithNullLabelsShouldFail() {
+        try {
+            testCall(db, "CALL apoc.merge.node([null, null], {name:'John'}) YIELD node RETURN node",
```
Lojjs, line 131: It would be good to also have a test where not all labels are null, like `CALL apoc.merge.node(['Person', null], {name:'John'})...`

core/src/main/java/apoc/merge/Merge.java
```diff
@@ -83,16 +82,20 @@ public Stream<NodeResultWithStats> nodeWithStats(@Name("labels") List<String> la
...
     private Result getNodeResult(List<String> labelNames, Map<String, Object> identProps, Map<String, Object> props, Map<String, Object> onMatchProps) {
-        if (identProps ==null || identProps.isEmpty()) {
+        if (identProps == null || identProps.isEmpty()) {
             throw new IllegalArgumentException("you need to supply at least one identifying property for a merge");
         }
 
-        String labels = labelString(labelNames);
+        if (labelNames != null && (labelNames.contains(null) || labelNames.contains(""))) {
+            throw new IllegalArgumentException("The list of label names may not contain any null or empty String values. If you wish to merge a node without a label, pass an empty list instead.");
+        }
```
Lojjs, line 91: I don't know if we want to use the new types here like NODE, LIST and STRING, maybe check with @gem-neo4j as she has worked hard to unify the use of them.

Same applies to the relationship error message

core/src/test/java/apoc/merge/MergeTest.java
```diff
@@ -136,23 +124,59 @@ public void testMergeWithEmptyIdentityPropertiesShouldFail() {
...
+        } catch (QueryExecutionException e) {
+            assertEquals(e.getMessage(), "Failed to invoke procedure `apoc.merge.node`: Caused by: java.lang.IllegalArgumentException: " +
+                    "The list of label names may not contain any null or empty String values. If you wish to merge a node without a label, pass an empty list instead.");
+        }
+    }
+
+    @Test
+    public void testMergeNodeContainingEmptyLabelShouldFail() {
+        try {
+            testCall(db, "CALL apoc.merge.node([''], {name:'John'}) YIELD node RETURN node",
```
Lojjs, line 155: This test looks identical to the one above, I guess that you intended to send in '' rather than [''] here?

core/src/test/java/apoc/merge/MergeTest.java
```diff
@@ -136,23 +124,59 @@ public void testMergeWithEmptyIdentityPropertiesShouldFail() {
...
+                    row -> assertTrue(row.get("node") instanceof Node));
+            fail();
+        } catch (QueryExecutionException e) {
+            assertEquals(e.getMessage(), "Failed to invoke procedure `apoc.merge.node`: Caused by: java.lang.IllegalArgumentException: " +
+                    "The list of label names may not contain any null or empty String values. If you wish to merge a node without a label, pass an empty list instead.");
+        }
+    }
+
+    @Test
+    public void testMergeNodeWithEmptyLabelListShouldFail() {
```
Lojjs, line 141: It would be good to also have a test where not all labels are empty, like CALL apoc.merge.node(['Person' , ''], {name:'John'})...

Lojjs (changes requested):
The check looks good but I have some comments on the tests and messages.

core/src/main/java/apoc/merge/Merge.java
gem-neo4j: I see the update and will add my comment here to continue the thread:

"The list of label names may not contain any `NULL` or empty `STRING` values. If you wish to merge a `NODE` without a label, pass an empty `LIST` instead."

The first list has been missed, and the second time list is mentioned, it mentions `LIST` by itself, which is not a valid Cypher Type name.

I am also unsure about our take on error messages containing personal pronouns?

Lojjs: There are some new guidelines for error messages but we will discuss those in a meeting in 35 minutes so they are not really official/well structured yet

Lojjs: This is WIP PR of error guidelines if you are curious 
https://github.com/neo4j/docs-status-codes/pull/73

nadja-muller: How about this?
"The list of label names may not contain any NULL or empty STRING values. If you wish to merge a NODE without a label, pass an empty list instead."

Lojjs: I think that is OK, do you agree @gem-neo4j ?

gem-neo4j: Yeah, I think that is fine :)

Lojjs (approved):
I think it is fine now

Lojjs (approved):
Looks fine with the renaming too

PR #504, opened 2023-10-04
[x2fDvuKu] extend timeout for flaky BigGraphTest

nadja-muller:
apoc.refactor.rename.type reuses functionality from periodic iterate, which uses a termination guard. However, there is no concrete guarantee on when we will catch it. I propose we extended the timemout to encompass the outlier in the build. The default value was 10L seconds.

When running the test locally, timepassed was 0. Doubles could be a bit more precise.


ncordon (approved):

nadja-muller:
Closing since the frequency of the test failure is less than 1 in a 1000

PR #508, opened 2023-10-09
[2kuFEKOe] Fix flaky test

loveleif:
- Add test case in `CypherInitializerTest` for using `CREATE DATABASE` in apoc initialize. This test case fails and seems to be a bug in the initializer.
- Fix flaky test in `TriggerEnterpriseFeaturesTest` by avoiding using initializer (revert once the initializer bug is fixed).


loveleif:
[Team city build](https://live.neo4j-build.io/buildConfiguration/CypherSurface_Apoc_Dev_BuildAndCreateArtifacts/17669479?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true)

PR #513, opened 2023-10-16
[5kItK998] Fix bug in triggers

loveleif:
Fix a bug where active triggers are not executed.

### DIFF
--- a/core/src/main/java/apoc/trigger/TriggerHandler.java
+++ b/core/src/main/java/apoc/trigger/TriggerHandler.java
@@ -20,6 +20,7 @@ package apoc.trigger;
 
 import static apoc.ApocConfig.APOC_TRIGGER_ENABLED;
 import static apoc.ApocConfig.apocConfig;
+import static apoc.SystemLabels.ApocTrigger;
 
 import apoc.ApocConfig;
 import apoc.Pools;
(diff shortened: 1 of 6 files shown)


core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -95,28 +98,31 @@ public void checkEnabled() {
     }
 
     private void updateCache() {
-        activeTriggers.clear();
```
loveleif, line 98: This was the problem, triggers were not run between this call and adding them all back later on.

core/src/main/java/apoc/trigger/TriggerHandler.java
```diff
@@ -57,15 +58,17 @@ private enum Phase {before, after, rollback, afterAsync}
...
+
     private final Log log;
     private final GraphDatabaseService db;
     private final DatabaseManagementService databaseManagementService;
     private final ApocConfig apocConfig;
     private final Pools pools;
     private final JobScheduler jobScheduler;
 
-    private long lastUpdate;
+    private volatile long lastUpdate;
```
ncordon, line 71: Why does this need to be thread safe?

ncordon (approved):
The traceability doesn't matter for merging the pr but the LICENSES need to be up to date and they depend on the neo4j version been run. You'll have to download the latest neo4j SNAPSHOT and do `./gradlew generateLicensesFiles`

core/src/main/java/apoc/trigger/TriggerHandler.java
loveleif: It can be set and read from multiple different threads at the same time (from scheduled thread and from procedures calling into here).

PR #520, opened 2023-11-01
[NOID] Adds job to apply formatting in PRs

ncordon:
## What
Adds a job to run formatting in PRs by writting `/spotlessApply` in it

## Why
Because especially for the licensing it's really annoying to have to update neo4j locally every time the license check fails 😢 . Doing a job for the licensing from this one is super easy, this has been done in a very reusable way. Most of the code is copied from the job we have in the monorepo.

PR #522, opened 2023-11-03
[NOID] Formats everything

ncordon:
Uses #520 

## Why

To have a common style everywhere


Lojjs (approved):
First commit looks fine. The second one have I not reviewed in detail ofc, but skimming through the first few files, it looks OK.

PR #524, opened 2023-11-03
[NOID] Adds job to update the licenses in PRs

ncordon:
## Why

So we don't have to manually download neo4j to apply new licenses


gem-neo4j (approved):
LGTM :)

PR #529, opened 2023-11-07
[jyX5Zi3L] Fix bug in apoc.search.nodeAll and apoc.search.node

loveleif:
`apoc.search.nodeAll` and `apoc.search.node` used to access a transaction from multiple threads.


gem-neo4j (approved):
LGTM! Thanks :D

PR #532, opened 2023-11-10
[NOID] Setups testing for JDK 21

ncordon:
## Why
We want to compile with JDK 17 but test with both JDK 17 and JDK 21

### DIFF
--- a/.github/actions/test-gradle-project/action.yaml
+++ b/.github/actions/test-gradle-project/action.yaml
@@ -4,11 +4,16 @@ inputs:
   project-name:
     description: "Name of gradle project to test"
     required: true
+  java-version:
+    description: "JDK to use"
+    required: true
 
 runs:
   using: "composite"
   steps:
     - uses: ./.github/actions/setup-jdk
+      with:
+        java-version: ${{inputs.java-version}}
     - uses: ./.github/actions/setup-gradle-cache
     - name: Run tests
       shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yaml
```diff
@@ -81,6 +92,7 @@ jobs:
       fail-fast: false
       matrix:
         project: ['common', 'core', 'processor', 'test-utils', 'it']
+        java-version: ['17', '21']
```
gem-neo4j, line 95: I remember there was a reason we had 17.0.8, but not what that reason was, is it no longer a problem?

.github/actions/test-gradle-project/action.yaml
```diff
@@ -4,15 +4,20 @@ inputs:
...
   using: "composite"
   steps:
     - uses: ./.github/actions/setup-jdk
+      with:
+        java-version: ${{inputs.java-version}}
     - uses: ./.github/actions/setup-gradle-cache
     - name: Run tests
       shell: bash
-      run: ./gradlew ${{inputs.project-name}}:check --parallel
+      run: ./gradlew ${{inputs.project-name}}:test --parallel
```
gem-neo4j, line 20: Is check not needed because we run the other things e.g spotless separately?

.github/workflows/CI.yaml
ncordon: Because we had to move it up specifically with a breaking change they did in JDK? But that shouldn't happen hopefully very often

.github/actions/test-gradle-project/action.yaml
ncordon: Yes. Because spotless didn't work in [f98d019](https://github.com/neo4j/apoc/pull/532/commits/f98d0198d5615d21652e6a7980cbf5461abc8033) for Java 21 🤷

gem-neo4j (approved):
LGTM :) Thank you

PR #537, opened 2023-11-16
[NOID] Use Injected UrlAccessChecker for checking access to URLs

mnd999:
Fixes #<Replace with the number of the issue, Mandatory>

Use Injected URLAccessChecker for checking access to URLs calling internal WebURLAccessRule directly
Proposed Changes (Mandatory)

Inject a new URLAccessChecker to separate out APOC from making internal calls into WebURLAccessRule, as this needs to changes to support RBAC permissions as well. Having this injectable will mean that we can late get this new functionality in APOC as well for free.

The alternative is to move the config only based checks into APOC.

### DIFF
--- a/core/src/main/java/apoc/export/json/ImportJson.java
+++ b/core/src/main/java/apoc/export/json/ImportJson.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.Scanner;
 import java.util.stream.Stream;
 import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.security.URLAccessChecker;
 import org.neo4j.procedure.Context;
 import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Mode;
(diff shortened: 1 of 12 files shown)


core/src/main/java/apoc/export/json/ImportJson.java
```diff
@@ -46,6 +47,9 @@ public class ImportJson {
     @Context
     public TerminationGuard terminationGuard;
 
+    @Context
+    public URLAccessChecker urlAccessChecker;
```
loveleif, line 51: I assume this breaks backwards compatibility, but that's ok perhaps.

loveleif (approved):
Nice, thanks!

core/src/main/java/apoc/export/json/ImportJson.java
mnd999: Yeah, it does. But that's okay, right - we require a matching APOC version with each Neo4j release? How would we usually go about merging something that breaks like this? Core first, then APOC?

PR #542, opened 2023-11-21
[G9u7OAg9] Ignore check for optimizations if the format is CSV

gem-neo4j:
CSV export can't use optimizations, so the check that it is correct is not very useful. In general we should probably clean up these configs as it is quite messy having all of them, but that requires an investigation into which procedure needs what 👀

### DIFF
--- a/common/src/main/java/apoc/export/util/ExportConfig.java
+++ b/common/src/main/java/apoc/export/util/ExportConfig.java
@@ -127,6 +127,10 @@ public class ExportConfig extends CompressionConfig {
     }
 
     public ExportConfig(Map<String, Object> config) {
+        this(config, ExportFormat.CYPHER_SHELL);
+    }
+
+    public ExportConfig(Map<String, Object> config, ExportFormat exportFormat) {
         super(config);
         config = config != null ? config : Collections.emptyMap();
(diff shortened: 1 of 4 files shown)


common/src/main/java/apoc/export/util/ExportConfig.java
```diff
@@ -169,7 +173,9 @@ private void validate() {
                     "`useOptimizations: 'UNWIND_BATCH_PARAMS'` can be used only in combination with `format: 'CYPHER_SHELL' but got [format:`"
                             + this.format + "]");
         }
-        if (!OptimizationType.NONE.equals(this.optimizationType) && this.unwindBatchSize > this.batchSize) {
+        // CSV doesn't use optimization type
+        if (!OptimizationType.NONE.equals(this.optimizationType) && this.unwindBatchSize > this.batchSize
+            && !ExportFormat.CSV.equals(this.format)) {
             throw new RuntimeException("`unwindBatchSize` must be <= `batchSize`, but got [unwindBatchSize:"
                     + unwindBatchSize + ", batchSize:" + batchSize + "]");
         }
```
ncordon, line 181: I'd say this message is not good. It's better to split into another condition where we throw a specific exception for the case where we've passed an optimization and a csv format?

ncordon (approved):

PR #543, opened 2023-11-24
[NOID] Quick Fix UtilIT to fix build

mnd999:
I will review this test again, we need to get away from depending on core internals, but let's just fix the build for now.


gem-neo4j (approved):
Seems fine to me :)

PR #544, opened 2023-11-24
[NOID] More fixes for issues related to URLAccessChecker

mnd999:
More fixes for issues related to URLAccessChecker

- Geocode does these checks as well, and that was missing.
- UtilIT rewritten again to mock URLAccessChecker so as to not rely on coredb internals


gem-neo4j (approved):
Seems fine to me :)

PR #547, opened 2023-11-29
[HTtgtRFX] Fix flaky stress test

loveleif:
- Fix bug in triggers, last update was too early.
- Fix flaky test assertion in `TriggerEnterpriseFeaturesTest.stressTest`. I can't find nothing wrong with the assertions, but they sometimes caused `EntityNotFoundException` during installing of triggers and hopefully this helps.
- Apply code style to the whole project.

### DIFF
--- a/it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
+++ b/it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
@@ -21,7 +21,7 @@ package apoc.it.core;
 import static apoc.ApocConfig.APOC_CONFIG_INITIALIZER;
 import static apoc.ApocConfig.APOC_TRIGGER_ENABLED;
 import static apoc.SystemPropertyKeys.database;
-import static apoc.it.core.CreateTriggers.CreateTrigger;
+import static apoc.it.core.CreateAndDropTriggers.CreateTrigger;
 import static apoc.trigger.TriggerHandler.TRIGGER_REFRESH;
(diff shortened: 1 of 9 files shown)


it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
```diff
@@ -435,32 +433,20 @@ public void stop() {
...
-            for (int i = 0; i < iterations; ++i) {
-                final var name = "temp-trigger-" + i;
-                session.run(CreateTrigger, Map.of("db", db, "name", name, "trigger", "RETURN 1"))
-                        .consume();
-            }
-        }
-    }
-}
-
-record DropRandomTriggers(Driver driver, String db, int iterations) implements Runnable {
```
loveleif, line 453: Note, before we dropped and triggers at random concurrently to creating them. I see no reason why that should not be possible. But since install trigger failed sometimes (EntityNotFound) I change it here to not drop and install concurrently.

gem-neo4j (approved):
This looks good to me :) Slightly confused by the test, but it wasn't your changes that were confusing at least

it/src/test/java/apoc/it/core/TriggerEnterpriseFeaturesTest.java
loveleif: It's not the clearest test, but it was really good at reproducing a previous bug. Your statements are correct. The randomness does not serve that much purpose now. I guess I kept it to have some triggers live on for different amounts of time in case that makes a difference.

PR #550, opened 2023-12-04
[NOID] Allows execution of integration tests with bundled apoc

ncordon:
## What

Allows execution of integration tests with bundled apoc rather than building a jar we plug passing a property to `gradlew`:

```
./gradlew -PtestDockerBundle  it:test
```

## Why
Because the release ops team wants to have a build to test the docker image we release.


test-utils/src/main/java/apoc/util/TestContainerUtil.java
```diff
@@ -129,24 +131,32 @@ private static Neo4jContainerExtension createNeo4jContainer(
...
+            }
+        }
 
-            copyFilesToPlugin(
-                    new File(projectDir, "build/libs"),
-                    new WildcardFileFilter(Arrays.asList("*-extended.jar", "*-core.jar")),
-                    pluginsFolder);
+        if (testDockerBundle && apocPackages.contains(ApocPackage.EXTENDED)) {
+            throw new IllegalArgumentException("You cannot run these tests with apoc extended bundled inside"
+                    + "the docker container because only apoc core comes bundled in those");
```
gem-neo4j, line 155:
```suggestion
            throw new IllegalArgumentException("You cannot run these tests with apoc extended bundled inside "
                    + "the docker container because only apoc core comes bundled in those");
```

gem-neo4j (approved):
LGTM, nice :)

PR #553, opened 2023-12-05
[NOID] Solves populated git-diff index bug when no changes have happened

ncordon:
## What
Solves bug in bot to format / update licenses where `git diff-index` could be returning results even if no files have had updates.

This bug probably is present in the monorepo as I copied those files from there.

## Why
For reference the answer to this question https://stackoverflow.com/questions/3878624/how-do-i-programmatically-determine-if-there-are-uncommitted-changes

> if there are files that have been touched, but whose contents are the same as in the index, you'll need to run git update-index --refresh before git diff-index, otherwise diff-index will incorrectly report that the tree is dirty)

### DIFF
--- a/.github/actions/gradle-command-on-pr/action.yaml
+++ b/.github/actions/gradle-command-on-pr/action.yaml
@@ -38,7 +38,9 @@ runs:
     - name: Check for modified files
       shell: bash
       id: git-check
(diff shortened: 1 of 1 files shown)


Lojjs (approved):
Nice catch, I guess as you are saying that we should do same thing to monorepo too

PR #560, opened 2023-12-27
[tyXNOmR9] Investigate flaky test

loveleif:
- Improve assertion to investigate flaky test, instead of saying "false was not equal to true" it will report the full stack trace and error message.
- Increase timeout.


AzuObs (approved):
Looks good to me! Great work

PR #561, opened 2024-01-04
[yaWCMgIP] Exclude asm to fix startup issue

gem-neo4j:
I tested this manually by copying the jar to desktop and changing the name to: apoc5plus-5.15.0, without this change it fails, with it works (also with it, there is no trace of it in the logs either)

### DIFF
--- a/LICENSES.txt
+++ b/LICENSES.txt
@@ -474,7 +474,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 ------------------------------------------------------------------------------
 BSD-3-Clause
-  asm-9.3.jar
   asm-9.6.jar
   asm-analysis-9.6.jar
   asm-commons-5.0.3.jar
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -281,7 +281,6 @@ BSD-2-Clause
   zstd-jni-1.5.5-10.jar
 
 BSD-3-Clause
-  asm-9.3.jar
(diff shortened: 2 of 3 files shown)


ncordon (changes requested):
Are we sure the dependency is `asm` is included in neo4j? 

What kernel said in their messages is that it was included in the `neo4j-jwt-addon.jar` plugin, which is not always present (in Desktop it seems it always is).

For this to work and not cause binary incompatibility issues on runtime we need to be sure `asm` is shipped with the community projects of neo4j.

gem-neo4j:
I think so, what I did was I checked the dependencies in APOC, and it was listed under org.neo4j:neo4j-codegen:5.16.0-SNAPSHOT, org.neo4j:test-utils:5.16.0-SNAPSHOT as well, and the POM in codegen for example was:

```
                <configuration>
                    <artifactSet>
                        <includes>
                            <include>org.ow2.asm:*</include>
                        </includes>
                    </artifactSet>
                    <relocations>
                        <relocation>
                            <pattern>org.objectweb.asm</pattern>
                            <shadedPattern>org.neo4j.codegen.asm</shadedPattern>
                        </relocation>
                    </relocations>
                    <createDependencyReducedPom>false</createDependencyReducedPom>
                </configuration>
```

I assumed include here implies it is shipped with Neo. Is there another way I can check?

neo-technology-build-agent:
:x: [generateLicenses failed][1]

[1]: https://github.com/neo4j/apoc/actions/runs/7420272863

ncordon (approved):

PR #569, opened 2024-01-10
[NOID] Use count methods with transaction state

zimmre:
Use count methods with tx state instead of ones without tx state that are deprecated and being removed.


ncordon (approved):

loveleif (approved):

loveleif:
Don't know why @ncordon assigned himself, but better wait for his review since I guess he might have some insight.

PR #573, opened 2024-01-23
[h7Q1epCi] Make hashing functions stable on all neo4j types

loveleif:
The hash functions are only intended for strings, but unfortunately their signature allows all types. This is a safety for users that use hash functions on non string values despite the documentation.

### DIFF
--- a/core/src/main/java/apoc/util/Utils.java
+++ b/core/src/main/java/apoc/util/Utils.java
@@ -18,9 +18,14 @@
  */
 package apoc.util;
 
+import static java.util.stream.Collectors.joining;
+import static org.apache.commons.codec.binary.StringUtils.getBytesUtf8;
+
+import java.security.MessageDigest;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
+import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.neo4j.graphdb.TransactionTerminatedException;
 import org.neo4j.procedure.*;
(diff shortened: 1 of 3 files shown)


core/src/test/java/apoc/util/UtilsTest.java
```diff
@@ -291,6 +302,129 @@ public void testMd5() {
...
+            long seed, Object val, Transaction tx, String query, Map<String, Object> params) {
+        final var a = tx.execute(query, params).stream().toList();
+        final var b = tx.execute(query, params).stream().toList();
+        final var message = "%s should have stable hash (seed=%s)".formatted(val, seed);
+        assertEquals(message, a, b);
+        assertEquals(1, a.size());
+        final var first = a.get(0).entrySet().iterator().next();
+        for (final var e : a.get(0).entrySet()) {
+            assertEquals(message, first.getValue(), e.getValue());
+        }
```
Lojjs, line 425: Can you explain the point of this assertion? To me it looks like you are comparing something to itself

core/src/test/java/apoc/util/UtilsTest.java
```diff
@@ -291,6 +302,129 @@ public void testMd5() {
...
+                """
+                            .formatted(hashFunc, listCypher, hashFunc);
+            assertStableHash(seed, randStorables, tx, listOfListsQuery, Map.of("value", randStorables));
+
+            final var nodesHashQuery =
+                """
+                match (a:HashFunctionsAreStable)-[r]->(b:HashFunctionsAreStable)
+                return
+                  apoc.util.%s([a, b, r, [a, b, r]]) as hash1,
+                  apoc.util.%s([a, b, r, [a, b, r]]) as hash2
```
Lojjs, line 407: Is it intentional the hash1 and hash2 is the same in this query? Otherwise, you could remove the hash2 line

core/src/main/java/apoc/util/Utils.java
```diff
@@ -134,4 +134,40 @@ public byte[] compress(
...
+    private static String hexHash(final MessageDigest digest, final List<Object> values) {
+        for (final var value : values) digest.update(getBytesUtf8(toHashString(value)));
+        return Hex.encodeHexString(digest.digest());
+    }
+
+    /*
+     * This is not the most efficient way to produce a hash, but it is backwards compatible.
+     * This function is only intended to be used on strings (as documented on the hash functions above)
+     * But it turns out that is not how everyone is using it, so as a safety we have stable implementations
+     * for all neo4j types.
```
Lojjs, line 147: Nice that you managed to keep backwards compatibility. Do you think we should mention something about the other types than strings in docs or not?

Lojjs (changes requested):
Nice work, just some minor questions and remarks.

core/src/test/java/apoc/util/UtilsTest.java
```diff
@@ -291,6 +302,129 @@ public void testMd5() {
...
+                .map(t -> rand.nextValueOfType(t).asObject())
+                .toList();
+        final var randMap =
+                range(0, randStorables.size()).boxed().collect(toUnmodifiableMap(i -> "p" + i, randStorables::get));
+
+        // Create node with random property values
+        try (final var tx = db.beginTx()) {
+            final var node = tx.createNode(Label.label("HashFunctionsAreStable"));
+            final var rel = node.createRelationshipTo(node, RelationshipType.withName("R"));
+            for (int i = randStorables.size() - 1; i >= 0; --i) {
```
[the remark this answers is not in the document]
loveleif: Yes! `{a: 1, b: 2}` should have the same hash as `{b: 2, a: 1}`, before we did toString, and that will simply use the iteration order (which depends on the map implementation). I wanted to try to make the properties come in another order (not sure it succeeded, depends on the map implementation and what kernel decides to do).

core/src/main/java/apoc/util/Utils.java
loveleif: Yes, perhaps we should point out that it is only meant for strings and other types can have bad performance?

core/src/test/java/apoc/util/UtilsTest.java
```diff
@@ -291,6 +302,129 @@ public void testMd5() {
...
+        hashIsStable("sha1");
+    }
+
+    @Test
+    public void sha256IsStableOnAllTypes() {
+        hashIsStable("sha256");
+    }
+
+    @Test
+    public void sha384IsStableOnAllTypes() {
```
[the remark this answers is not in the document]
loveleif: Thanks!!

core/src/test/java/apoc/util/UtilsTest.java
loveleif: I can, it's a bit messy. Let's look at an example:

We might run this query with `$value=23, $list=List(23)` and `n.p1=23`:

```
match (n)
return
  apoc.util.md5([n.p1]) as hash1,
  apoc.util.md5([$value]) as hash2
  apoc.util.md5($list) as hash3
```

This assertion is intended to make sure `hash1`, `hash2` and `hash3` have the same value. They should all equal each other, but in Cypher there can be subtle differences depending on where a value comes from. For example `create (n {p: [1,2,3]}) return n.p` will most probably not have exactly the same types internally as `return [1,2,3]` even though the two lists are equal and represents the same thing. Just wanted to make sure that such internal types do not affect the hash.

core/src/test/java/apoc/util/UtilsTest.java
loveleif: It can be removed, it's pretty useless. :)

core/src/test/java/apoc/util/UtilsTest.java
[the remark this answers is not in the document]
loveleif: Added it

core/src/test/java/apoc/util/UtilsTest.java
loveleif: Removed

core/src/test/java/apoc/util/UtilsTest.java
Lojjs: I see, thanks for explaining!

Lojjs (approved):
Looks good now

PR #574, opened 2024-01-24
[NOID] Fixes directory that should be non existing

ncordon:
## What
Changes the name of the directory `apoc` to `"__non-existing-dir__"` so we get more chances of that directory truly not existing.

## Why
Because in some internal build we have an existing `apoc` folder which was making the tests not work because they expect that folder to not exist.

### DIFF
--- a/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
+++ b/core/src/test/java/apoc/export/ExportCoreSecurityTest.java
@@ -174,14 +174,18 @@ public class ExportCoreSecurityTest {
         not found. They all attempt to exit the import folder back to the apoc folder:
         Directory Layout: .../apoc/core/target/import
         */
-        private static final String case10 = "file://%2e%2e%2f%2e%2e%2f%2e%2e%2f%2e%2e%2f/apoc/" + FILENAME;
-        private static final String case11 = "file://../../../../apoc/" + FILENAME;
-        private static final String case12 = "file:///..//..//..//..//apoc//core//..//" + FILENAME;
-        private static final String case13 = "file:///..//..//..//..//apoc/" + FILENAME;
-        private static final String case14 =
-                "file://" + directory.getAbsolutePath() + "//..//..//..//..//apoc/" + FILENAME;
-        private static final String case15 =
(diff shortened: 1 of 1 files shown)


core/src/test/java/apoc/export/ExportCoreSecurityTest.java
```diff
@@ -174,14 +174,18 @@ public TestIllegalExternalFSAccess(
...
         */
-        private static final String case10 = "file://%2e%2e%2f%2e%2e%2f%2e%2e%2f%2e%2e%2f/apoc/" + FILENAME;
-        private static final String case11 = "file://../../../../apoc/" + FILENAME;
-        private static final String case12 = "file:///..//..//..//..//apoc//core//..//" + FILENAME;
-        private static final String case13 = "file:///..//..//..//..//apoc/" + FILENAME;
-        private static final String case14 =
-                "file://" + directory.getAbsolutePath() + "//..//..//..//..//apoc/" + FILENAME;
-        private static final String case15 =
-                "file:///%252e%252e%252f%252e%252e%252f%252e%252e%252f%252e%252e%252f/apoc/" + FILENAME;
+        private static final String nonExistingDirectory = "__non-existing-dir__";
```
loveleif, line 177: We could use `Files.createTempDirectory` and then we don't have to worry about leaving stuff around on the host running the build.

loveleif (approved):
Nice catch!!

core/src/test/java/apoc/export/ExportCoreSecurityTest.java
ncordon: But it's not going to create the file. If we do create the directory (even if it's temporary) then the tests would fail. I could make the name random if you prefer

loveleif: Ahh, ok! I thought it would create some files there at some point. Because if it does we could do `Files.createTempDirectory().resolve("non-existing")` and have a non existing directory and be sure that those files will not remain on the host. But if we don't it doesn't matter. Feel free to merge!

PR #578, opened 2024-01-26
[f3ZUxEG2] Fix bugs and refactor apoc.cypher.runMany

loveleif:
Fixes two issues in `apoc.cypher.runMany`:

- Transactions were accessed from multiple threads, this is not allowed and has undefined behavior (this has always been the case but recently this has started to cause more problems like flaky tests, probably some change in kernel that makes it more sensitive).
- Errors silently stopped the procedure from processing. This was not documented so I assume it was an oversight.

For simplicity, the implementation is changed to be single threaded. We previously spawned a second thread for executing inner queries. While this could have performance benefits in some situations, there is also an overhead of starting and synchronizing threads. We currently have no benchmarks for this procedure so there's not much to base this decision on either way.

Some undocumented behavior is kept from previous implementation:

- Ignore `AuthorizationViolationException` errors silently.
- Run inner statements in new transactions.

### DIFF
--- a/core/src/main/java/apoc/cypher/Cypher.java
+++ b/core/src/main/java/apoc/cypher/Cypher.java
@@ -18,6 +18,7 @@
  */
 package apoc.cypher;
 
+import static apoc.cypher.Cypher.toMap;
 import static apoc.cypher.CypherUtils.runCypherQuery;
 import static apoc.cypher.CypherUtils.withParamMapping;
 import static apoc.util.MapUtil.map;
@@ -27,18 +28,15 @@ import static org.neo4j.procedure.Mode.WRITE;
 
 import apoc.Pools;
 import apoc.result.MapResult;
-import apoc.util.EntityUtil;
-import apoc.util.QueueBasedSpliterator;
 import apoc.util.Util;
-import java.io.Reader;
+import apoc.util.collection.Iterators;
 import java.io.StringReader;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Scanner;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
+import java.util.Spliterator;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -48,6 +46,7 @@ import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.QueryStatistics;
 import org.neo4j.graphdb.Result;
 import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.security.AuthorizationViolationException;
 import org.neo4j.procedure.Context;
 import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Mode;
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -81,103 +79,31 @@ public Stream<MapResult> run(@Name("statement") String statement, @Name("params"
...
-    private Stream<RowResult> runManyStatements(
-            Reader reader,
-            Map<String, Object> params,
-            boolean schemaOperation,
-            boolean addStatistics,
-            int queueCapacity) {
-        BlockingQueue<RowResult> queue = runInSeparateThreadAndSendTombstone(
-                queueCapacity,
-                internalQueue -> {
-                    if (schemaOperation) {
```
loveleif, line 93: This was always false ¯\_(ツ)_/¯

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -81,103 +79,31 @@ public Stream<MapResult> run(@Name("statement") String statement, @Name("params"
...
-    }
-
-    private void runDataStatementsInTx(
-            Reader reader, BlockingQueue<RowResult> queue, Map<String, Object> params, boolean addStatistics) {
-        Scanner scanner = new Scanner(reader);
-        scanner.useDelimiter(";\r?\n");
-        while (scanner.hasNext()) {
-            String stmt = removeShellControlCommands(scanner.next());
-            if (stmt.trim().isEmpty()) continue;
-            if (!isSchemaOperation(stmt)) {
```
loveleif, line 136: Here we used to filter out schema operations, they were simply never run. But why?

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -193,26 +119,7 @@ public Stream<RowResult> runManyReadOnly(
     private static final Pattern shellControl =
             Pattern.compile("^:?\\b(begin|commit|rollback)\\b", Pattern.CASE_INSENSITIVE);
 
-    private Object consumeResult(Result result, BlockingQueue<RowResult> queue, boolean addStatistics) {
-        try {
-            long time = System.currentTimeMillis();
-            int row = 0;
-            while (result.hasNext()) {
-                terminationGuard.check();
-                Map<String, Object> mapResult = EntityUtil.anyRebind(tx, result.next());
```
loveleif, line 202: Here we used to access `tx` from a different thread which is not allowed.

core/src/test/java/apoc/export/cypher/ExportCypherMultiRelTest.java
```diff
@@ -93,74 +91,63 @@ public void teardown() {
 
     @Test
     public void updateAllOptimizationNone() {
-        String expectedCypherStatement =
-                NODES_MULTI_RELS + SCHEMA_WITH_UNIQUE_IMPORT_ID + RELS_MULTI_RELS + CLEANUP_SMALL_BATCH;
+        String expectedCypherStatement = NODES_MULTI_RELS + RELS_MULTI_RELS + CLEANUP_SMALL_BATCH;
         final Map<String, Object> map = withoutOptimization(map("cypherFormat", "updateAll"));
 
         testsCommon(expectedCypherStatement, map);
     }
```
loveleif, line 98: These tests were weird. They used to have schema modification queries in them that are silently skipped in `runMany`.

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -222,15 +143,7 @@ private String removeShellControlCommands(String stmt) {
         return stmt;
     }
 
-    private boolean isSchemaOperation(String stmt) {
-        return stmt.matches("(?is).*(create|drop)\\s+(index|constraint).*");
-    }
-
-    private boolean isPeriodicOperation(String stmt) {
-        return stmt.matches("(?is).*using\\s+periodic.*");
-    }
```
loveleif, line 231: Regular expression based cypher parsers 😅

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -81,103 +80,46 @@ public Stream<MapResult> run(@Name("statement") String statement, @Name("params"
...
-                        try (Result result = threadTx.execute(stmt, params)) {
-                            return consumeResult(result, queue, addStatistics);
-                        }
-                    });
-                }
-            }
-        }
-    }
-
-    private void runSchemaStatementsInTx(
```
loveleif, line 153: This was never reached.

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -359,3 +265,51 @@ public Stream<MapResult> doWhenCase(
...
+    private int rowCount;
+
+    RunManyResultSpliterator(Result result, boolean statistics) {
+        this.result = result;
+        this.start = System.currentTimeMillis();
+        this.statistics = statistics;
+    }
+
+    @Override
+    public boolean tryAdvance(Consumer<? super Cypher.RowResult> action) {
```
loveleif, line 282: We could theoretically get better performance by overriding `forEachRemaining` using `Result.accept(ResultVisitor)`. But I believe procedure framework will traverse the stream one by one anyway so we could not get those benefits without changes there too.

core/src/main/java/apoc/cypher/Cypher.java
```diff
@@ -81,103 +80,46 @@ public Stream<MapResult> run(@Name("statement") String statement, @Name("params"
...
+            // It has been written, in some long forgotten commits,
+            // that failures of this kind should be avoided. The ancestors
+            // were brave and used a regex based cypher parser to avoid
+            // trying to execute schema changing statements all together.
+            // We don't have that courage, and try to forget about it
+            // after the fact instead.
+            // One can only hope that by keeping this tradition alive,
+            // in some form, we make some poor souls happier.
+            innerTx.close();
+            return Stream.empty();
```
loveleif, line 118: I was very tempted to not include this, but decided to anyway, many tests relies on this.

gem-neo4j (approved):
I think this looks like a good improvement for sure! The silent ignoring of schema operations is weird, but APOC is weird so 🤷‍♀️

core/src/main/java/apoc/cypher/Cypher.java
loveleif: Sounds good, I'll keep it here to not break tests and create a card to investigate if this is really what we want to do.

loveleif: https://trello.com/c/5iHK0d0X

PR #579, opened 2024-01-26
[NOID] Upgrade driver used in tests to 5.16.0

gjmwoods:
Updates the java-driver dependency used in the ITs to the latest currently (from `4.4.12` -> `5.16.0`).

There have been some [additional checks](https://neo4j.com/docs/java-manual/current/session-api/#java-driver-simple-result-consume) added in the 5.x series of the driver. This PR also migrates some usages of the driver from the deprecated `writeTransaction()` to `executeWrite()`.


loveleif (approved):
Thanks!! I've been wanting to do this for a while.

PR #581, opened 2024-02-01
[7nVhdJu3] apoc.convert.toTree keep all rels in tree fix

gem-neo4j:
Adds element id and rel element id to the tree, this allows us to check if a relationship-node pair is unique, before the assumption was that it was unique if the rel type and node id pair was unique, which meant nodes with 2 rels of the same type, pointing to the same node were ignored.

These extra fields are additive, which is why I think it is okay, let me know if you think this is breaking or not!

### DIFF
--- a/core/src/main/java/apoc/convert/Json.java
+++ b/core/src/main/java/apoc/convert/Json.java
@@ -170,14 +170,12 @@ public class Json {
     }
 
     @Procedure("apoc.convert.toTree")
-    @Description(
-            "Returns a stream of `MAP` values, representing the given `PATH` values as a tree with at least one root.")
-    // todo optinally provide root node
+    @Description("Returns a stream of `MAP` values, representing the given `PATH` values as a tree with at least one root.")
     public Stream<MapResult> toTree(
             @Name("paths") List<Path> paths,
             @Name(value = "lowerCaseRels", defaultValue = "true") boolean lowerCaseRels,
             @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
-        if (paths.isEmpty()) return Stream.of(new MapResult(Collections.emptyMap()));
+        if (paths == null || paths.isEmpty()) return Stream.of(new MapResult(Collections.emptyMap()));
         ConvertConfig conf = new ConvertConfig(config);
         Map<String, List<String>> nodes = conf.getNodes();
         Map<String, List<String>> rels = conf.getRels();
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/convert/ConvertJsonTest.java
```diff
@@ -782,14 +839,57 @@ public void testToTreeLeafNodesWithConfigErrorInclude() {
...
+                    Map<String, Object> rel1 = Map.of("p2", "node22", "name", "Node 22", "id", "n22", "R.prop1", "n21->n22 [1]", "R.prop2", "Rel 1");
+                    Map<String, Object> rel2 = Map.of("p2", "node22", "name", "Node 22", "id", "n22", "R.prop1", "n21->n22 [2]", "R.prop2", "Rel 2");
+                    assertEquals(2, RBranches.size());
+                    if (RBranches.get(0).entrySet().containsAll(rel1.entrySet())) {
+                        assertTrue(RBranches.get(1).entrySet().containsAll(rel2.entrySet()));
+                    } else {
+                        assertTrue(RBranches.get(0).entrySet().containsAll(rel1.entrySet()));
+                        assertTrue(RBranches.get(1).entrySet().containsAll(rel2.entrySet()));
+                    }
+                    assertFalse(res.hasNext());
```
loveleif, line 880: I have recently learned to appreciate `org.assertj.core.api.Assertions.assertThat` assertions (heavily used in the java parts of the mono repo). Their API is very vast, and it's easy to get lost, https://assertj.github.io/doc/#assertj-overview gives a quick overview what it is about.

The main benefit, for me, is improved error messages, if you don't want to use the full extent of the API it is often useful at least wrap the existing assertions. In this example that could look like for example:

```java
org.assertj.core.api.Assertions.assertThat(res.stream().toList()).satisfiesExactly(row -> {
    Map<String, Object> value = (Map<String, Object>) row.get("value");
    // The existing assertions, except assertFalse(res.hasNext()); (not needed because satisfiesExactly)
});
```

If this fails it will not only tell you `x was not equal to y`, but give you the complete context `Expected all entries in <the complete result> to satisfy all requirements but failed with: x was not equal to y`. Super helpful when tests start failing because you get the complete picture.

Anyway, I don't expect you to change(!), but if that sounds nice, you could consider to use assertj at some other time :).

loveleif (approved):
Great! Have a nice weekend!

core/src/main/java/apoc/convert/Json.java
```diff
@@ -203,8 +201,12 @@ public Stream<MapResult> toTree(
                     // parent-[:HAS_CHILD]->(child) vs. (parent)<-[:PARENT_OF]-(child)
                     if (!nMap.containsKey(typeName)) nMap.put(typeName, new ArrayList<>(16));
                     List<Map<String, Object>> list = (List) nMap.get(typeName);
+                    // Check that this combination of rel and node doesn't already exist
                     Optional<Map<String, Object>> optMap = list.stream()
-                            .filter(elem -> elem.get("_id").equals(m.getId()))
+                            .filter(elem ->
+                                    elem.get("_elementId").equals(m.getElementId())
+                                            && elem.get(typeName + "._elementId").equals(r.getElementId())
```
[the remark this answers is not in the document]
gem-neo4j: Yep, exactly :)

gem-neo4j:
Docs: https://github.com/neo4j/docs-apoc/pull/254

core/src/test/java/apoc/convert/ConvertJsonTest.java
gem-neo4j: Ah this was interesting, I will keep in mind for the next APOC task! :D

Issue #583, opened 2024-02-07
apoc.meta.data() fails when there is a relationship-type `RELATIONSHIP` breaks GenAI integrations


jexp:
When you have a graph with a `RELATIONSHIP` the current apoc.meta.data() implementation breaks with this exception.

Failed to invoke procedure `apoc.meta.data`: Caused by: java.lang.IllegalArgumentException: duplicate element: RELATIONSHIP

As we use this in the schema generation for our genai-integrations, this breaks them and the applications using them.

kamzz1107:
same issue I have faced today while calling apoc.meta.data()

mfreeman451:
this was marked completed and closed, but its not clear why? where is the fix? I'm having this same issue.

mfreeman451:
is there an earlier release that doesn't have this bug?

gem-neo4j:
The linked PR request fixes it, but the fix itself will be released in 5.18 :)

mfreeman451:
> The linked PR request fixes it, but the fix itself will be released in 5.18 :)

What is the release cadence for neo4j?

PR #584, opened 2024-02-08
[wbVWvi0y] Solves apoc.meta.data bug with reserved names for relationships and nodes

ncordon:
Solves #583 
## Why

Because a node named `NODE` or a relationship named `RELATIONSHIP` could not be used.

When doing

```
CREATE (a:NODE)-[r:RELATIONSHIP]->(m:Movie)
CALL apoc.meta.data()
```

we were getting:

```
java.lang.IllegalArgumentException: duplicate element: RELATIONSHIP
```

### DIFF
--- a/core/src/main/java/apoc/meta/Meta.java
+++ b/core/src/main/java/apoc/meta/Meta.java
@@ -84,6 +84,30 @@ import org.neo4j.procedure.UserFunction;
  * It is part of the APOC (Awesome Procedures on Cypher) library.
  */
 public class Meta {
+    private class MetadataKey {
+        Types type;
+        String key;
+
+        MetadataKey(Types type, String key) {
+            this.type = type;
+            this.key = key;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/meta/Meta.java
```diff
@@ -84,6 +84,30 @@
...
+            }
+
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            return (type.toString() + "-" + key).hashCode();
+        }
+    }
```
loveleif, line 110:
```suggestion
    private record MetadataKey(Types type, String key) {}
```

loveleif (approved):
Sweet!

core/src/main/java/apoc/meta/Meta.java
ncordon: Nice, didn't know this were in Java

gem-neo4j:
@ncordon Was this cherry-picked to 4.4 as well?

PR #585, opened 2024-02-09
[NOID] Does some CI maintenance

ncordon:
* Fixes spotless check
* Fixes deprecated node actions
* Makes archived test list name depend on the JDK (otherwise test results get overridden depending on whether it's JDK 17 or 21 that finishes first)
* Reformats some files

### DIFF
--- a/core/src/test/java/apoc/util/UtilsTest.java
+++ b/core/src/test/java/apoc/util/UtilsTest.java
@@ -405,11 +405,12 @@ public class UtilsTest {
             assertStableHash(seed, randStorables, tx, listOfListsQuery, Map.of("value", randStorables));
 
             final var nodesHashQuery =
-                """
+                    """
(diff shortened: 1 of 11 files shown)


core/src/test/java/apoc/util/UtilsTest.java
```diff
@@ -405,11 +405,12 @@ public void hashIsStable(String hashFunc) {
...
 
             final var nodesHashQuery =
-                """
+                    """
                 match (a:HashFunctionsAreStable)-[r]->(b:HashFunctionsAreStable)
                 return
                   apoc.util.%s([a, b, r, [a, b, r]]) as hash1
-                """.formatted(hashFunc, hashFunc);
+                """
+                            .formatted(hashFunc, hashFunc);
```
loveleif, line 413: The code formatter is weird in these cases.

loveleif (approved):
✨

gem-neo4j:
Closing as this is old and was never merged and the CI now looks quite different :)

PR #586, opened 2024-02-19
[4Bece3wd] Use Neo4j home to determine default config location.

Lojjs:
On Windows service, 'sun.java.command' is not supported. The fallback default path to the config folder was set to '.' which is not the actual default. With this solution, the home folder of Neo4j is instead used to calculate the default.

In order to get it to work when both APOC Core and Extended is installed, I reckon the same change should also be applied to [extended/src/main/java/apoc/ExtendedApocConfig.java](https://github.com/neo4j-contrib/neo4j-apoc-procedures/blob/dev/extended/src/main/java/apoc/ExtendedApocConfig.java)


loveleif (approved):
Nice!

PR #589, opened 2024-02-19
[8cS8e3IT] Optimise apoc.meta.* procedures

loveleif:
Optimize `apoc.meta.*` by using internal kernel APIs instead of inner queries.

657x speedup 💨 in the added benchmark (from 2 ops/s to 1314 ops/s).


core/src/main/java/apoc/convert/Json.java
```diff
@@ -272,7 +271,8 @@ private Map<String, Object> toMap(Node n, Map<String, List<String>> nodeFilters)
         Map<String, Object> props = n.getAllProperties();
         Map<String, Object> result = new LinkedHashMap<>(props.size() + 2);
         String type = Util.labelString(n);
-        result.put("_id", n.getId());;
+        result.put("_id", n.getId());
+        ;
```
gem-neo4j, line 275: I guess this `;` isn't needed?

settings.gradle
```diff
@@ -4,4 +4,5 @@ include("processor")
 include("common")
 include("core")
 include("test-utils")
-include("it")
\ No newline at end of file
+include("it")
+// include("benchmark")
```
gem-neo4j, line 8: Why is this commented out?

gem-neo4j (approved):
Looks good to me, nice to have a benchmark in APOC! I am not very knowledgable about benchmarks, did you just run it locally? or how did you test it? :)

settings.gradle
loveleif: I only ran the benchmark locally to get some numbers and since we don't run benchmarks on any server I thought it would be wasted to compile it.

loveleif:
> I am not very knowledgable about benchmarks, did you just run it locally? or how did you test it? :)

Yes, ran it locally (`./gradlew jmh`)

settings.gradle
gem-neo4j: Ah okay, maybe just add that as a comment then :)

PR #592, opened 2024-02-20
[NOID] Revert "[NOID] Bump arrow (#576)"

loveleif:
This reverts commit 128db22099f09446e4dfaf8f7a9207d4d958da78.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -41,13 +41,9 @@ Apache-2.0
   accessors-smart-2.5.0.jar
   annotations-17.0.0.jar
   arrow-format-13.0.0.jar
-  arrow-format-15.0.0.jar
   arrow-memory-core-13.0.0.jar
-  arrow-memory-core-15.0.0.jar
   arrow-memory-netty-13.0.0.jar
-  arrow-memory-netty-15.0.0.jar
   arrow-vector-13.0.0.jar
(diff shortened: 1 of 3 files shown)


NOTICE.txt
```diff
@@ -41,13 +41,9 @@ Apache-2.0
   accessors-smart-2.5.0.jar
   annotations-17.0.0.jar
   arrow-format-13.0.0.jar
-  arrow-format-15.0.0.jar
   arrow-memory-core-13.0.0.jar
-  arrow-memory-core-15.0.0.jar
   arrow-memory-netty-13.0.0.jar
-  arrow-memory-netty-15.0.0.jar
   arrow-vector-13.0.0.jar
-  arrow-vector-15.0.0.jar
```
MishaDemianenko, line 50: does this mean that before we were shipping both 13 and 15 arrow?

loveleif: I was also confused about this, but I assume so yes.

MishaDemianenko: but that would be a problem and a source of ooms in that case (if we would assume that its not plain arrow by default doing really strange things). Also in that preview i can see that we have the duplication of other libs as well

loveleif: Fixed here https://github.com/neo4j/apoc/pull/593

PR #594, opened 2024-02-20
[NOID] Align Netty dependency

loveleif:


Lojjs (approved):
Looks like this should help

PR #596, opened 2024-02-21
[2oZpORer] Fix illegal use of inner transaction in apoc.periodic.iterate

loveleif:
Make apoc.periodic.iterate rebind values before using them in inner transactions. Failing to do this can cause dead locks, failures and undefiened behaviour if the value contains nodes or relationships (because they carry transactions).


AzuObs (approved):
Great job 😄

PR #598, opened 2024-02-22
[NOID] Remove commons math3

klaren:
We only used a few methods from the library. Also, the `FastMath` that the library is built on claims to be faster and more accurate, but this is a lie in modern versions of Java. Performance will actually be worse since no intrinsics will be used.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -39,7 +39,6 @@ dependencies {
             strictly '3.14.0'
         }
     }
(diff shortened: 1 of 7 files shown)


loveleif (approved):
Sweet! Thanks, feel free to merge when builds pass.

klaren:
Replaced by https://github.com/neo4j/apoc/pull/601 since running CI on forks does not work.

PR #604, opened 2024-02-26
[NOID] Rely on neo4j provided libraries instead of shading

klaren:


loveleif (approved):
Thanks for cleaning this up!! Feel free to merge if build passes.

PR #605, opened 2024-02-26
[oIVz2Y7V] apoc.refactor.cloneNodes does not need to be in its own transaction

gem-neo4j:
There was a github issue where a user complained that if the Cypher query failed after this procedure ran, the procedure itself still committed the result. This is because it opened new transactions inside, this seemed overkill, so removed that from happening.

### DIFF
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -168,6 +168,19 @@ public class Util {
         return new Label[] {Label.label(labelNames.toString())};
     }
 
+    public static Label[] getLabelsArray(Node node) {
+        List<Label> labels = getLabels(node);
+        return labels.toArray(new Label[0]);
+    }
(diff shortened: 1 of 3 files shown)


common/src/main/java/apoc/util/Util.java
```diff
@@ -168,6 +168,19 @@ public static Label[] labels(Object labelNames) {
         return new Label[] {Label.label(labelNames.toString())};
     }
 
+    public static Label[] getLabelsArray(Node node) {
+        List<Label> labels = getLabels(node);
+        return labels.toArray(new Label[0]);
```
loveleif, line 173: You can go directly from stream to array

```suggestion
        return getLabelsAsStream(node).toArray(i -> new Label[i]);
```

loveleif (approved):
Great! 👌

PR #606, opened 2024-02-28
[NOID] Exclude OpenCSV bean-dependencies

klaren:
OpenCSV has two variants, one of which parses CSV into beans. We never use this so it's safe to exclude those dependencies.


loveleif (approved):

Lojjs (approved):
I think this looks alright, you have not touched the same parts of ApocConfig that I did in my support case last week

klaren:
OpenCSV has two variants, one of which parses CSV into beans. We never use this so it's safe to exclude those dependencies.

PR #607, opened 2024-03-04
[RHMU1hOu] Add tests for RBAC deny on load

gem-neo4j:
Add tests for all loading procedures in APAC to check the deny works well with RBAC rules.

docs: https://github.com/neo4j/docs-apoc/pull/260

### DIFF
--- a/it/src/test/java/apoc/it/core/RBACOnLoadTest.java
+++ b/it/src/test/java/apoc/it/core/RBACOnLoadTest.java
@@ -23,6 +23,7 @@ import static apoc.util.TestContainerUtil.createEnterpriseDB;
 import static apoc.util.TestContainerUtil.testCall;
 import static apoc.util.TestContainerUtil.testCallEmpty;
 import static java.util.Collections.emptyMap;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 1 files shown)


it/src/test/java/apoc/it/core/RBACOnLoadTest.java
```diff
@@ -0,0 +1,158 @@
...
+        testUserDriver = GraphDatabase.driver(neo4jContainer.getBoltUrl(), AuthTokens.basic(user, userP));
+        testUserDriver.verifyConnectivity();
+        testUserSession = testUserDriver.session();
+    }
+
+    private static void setupUser() {
+        List<String> queries = List.of(
+                "CREATE ROLE test",
+                "CREATE USER " + user + " SET PASSWORD '" + userP + "' SET PASSWORD CHANGE NOT REQUIRED",
+                "GRANT ROLE test, reader, editor TO testUser");
```
Lojjs, line 75: Reader is a subset of editor, so not sure it provides much value to explicitly grant the reader role

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+        testCallEmpty(session, query, emptyMap());
+    }
+
+    @Test
+    public void testRBACOnDeny() throws IOException {
+        String url = "https://neo4j.com/docs/cypher-refcard/3.3/csv/artists.csv";
+        addRBACOnLoad(url);
+
+        List<String> loadableAPOCProcs = List.of(
```
Lojjs, line 92: I looked at the docs for apoc.load.* and apoc.import.* to compare with this list, and there also seems to be apoc.load.arrow.stream

Lojjs (changes requested):
I think it looks good apart from two minor improvements that can be done. And kudos to @mnd999 for doing the feature work for us 🥇

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
gem-neo4j: apoc.load.arrow.stream takes a byte array and not a url :)

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
gem-neo4j: Ah yes, I added these at separate parts in my testing and didn't realise! thanks :D

Lojjs (approved):
Looks good now

PR #610, opened 2024-03-14
[olbYUBeG] Added meta-data to batches in periodic iterate

JoelBergstrand:
Transfers the metadata of the outer transaction to the inner transactions of the periodic.iterate procedure.

### DIFF
--- a/core/src/main/java/apoc/periodic/Periodic.java
+++ b/core/src/main/java/apoc/periodic/Periodic.java
@@ -43,6 +43,7 @@ import org.neo4j.graphdb.Transaction;
 import org.neo4j.graphdb.schema.ConstraintDefinition;
 import org.neo4j.graphdb.schema.IndexDefinition;
 import org.neo4j.graphdb.schema.Schema;
+import org.neo4j.kernel.impl.coreapi.InternalTransaction;
 import org.neo4j.logging.Log;
 import org.neo4j.procedure.*;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/apoc/periodic/Periodic.java
```diff
@@ -307,6 +308,13 @@ public Stream<BatchAndTotalResult> iterate(
                 "retries", 0)); // todo sleep/delay or push to end of batch to try again or immediate ?
         int failedParams = Util.toInteger(config.getOrDefault("failedParams", -1));
 
+        final Map<String, Object> metaData;
+        if(tx instanceof InternalTransaction iTx){
```
loveleif, line 312: Code style not applied (run `./gradlew spotlessApply`).

core/src/main/java/apoc/periodic/Periodic.java
```diff
@@ -333,6 +341,9 @@ public Stream<BatchAndTotalResult> iterate(
                     retries,
                     result,
                     (tx, p) -> {
+                        if(tx instanceof InternalTransaction iTx){
+                            iTx.setMetaData(metaData);
+                        }
```
loveleif, line 346: Perhaps we should do this for all procedures that runs inner queries? Create a card for it at least?

loveleif (approved):
Sweet!

PR #613, opened 2024-03-28
[z9rBzmVK] Make recursiveRebind nullsafe for maps.

Lojjs:
For maps which in turn contain maps or lists, we would call Collectors.toUnmodifiableMap, which throws a null pointer if the map also contains null values. This should be valid in Cypher.

Inspiration for the fix comes from https://stackoverflow.com/questions/24630963/nullpointerexception-in-collectors-tomap-with-null-entry-values

### DIFF
--- a/core/src/test/java/apoc/periodic/PeriodicTest.java
+++ b/core/src/test/java/apoc/periodic/PeriodicTest.java
@@ -720,7 +720,7 @@ public class PeriodicTest {
                 db,
                 """
                 CALL apoc.periodic.iterate(
-                    "WITH {a: null, b: {c: 1}} as value RETURN value",
+                    "WITH {a: null, b: {c: 1}, c: [null]} as value RETURN value",
                     "RETURN 1",
                      {})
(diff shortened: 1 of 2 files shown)


core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -714,6 +714,24 @@ public void testIterateFail() {
...
 
+    @Test
+    public void testIterateWithNullRebind() {
+        testResult(
+                db,
+                """
+                CALL apoc.periodic.iterate(
+                    "WITH {a: null, b: {c: 1}} as value RETURN value",
+                    "RETURN 1",
+                     {})
```
loveleif, line 725:
```suggestion
                CALL apoc.periodic.iterate(
                    "WITH {a: null, b: {c: 1}, c: [null]} as value RETURN value",
                    "RETURN 1",
                     {})
```

loveleif (approved):
Excellent!

PR #620, opened 2024-05-07
[NOID] Close resources for open readers

gem-neo4j:
Reports that the file is being reported as in use after the procedure has been used.

### DIFF
--- a/core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
+++ b/core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
@@ -336,6 +336,7 @@ public class XmlGraphMLReader {
             tx.rollback();
             throw e;
         } finally {
+            reader.close();
             tx.close();
         }
         return count;
--- a/core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
+++ b/core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
@@ -336,8 +336,8 @@ public class XmlGraphMLReader {
             tx.rollback();
             throw e;
         } finally {
-            reader.close();
             tx.close();
+            reader.close();
         }
         return count;
     }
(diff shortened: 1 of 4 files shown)


core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
```diff
@@ -336,6 +336,7 @@ public long parseXML(Reader input, TerminationGuard terminationGuard) throws XML
             tx.rollback();
             throw e;
         } finally {
+            reader.close();
```
loveleif, line 339: If this call throws an exception we will not call `tx.close()`. Might not be an issue in this case though, if I remember correctly transactions are closed on commit/rollback anyway.

loveleif (approved):
👍

core/src/main/java/apoc/export/graphml/XmlGraphMLReader.java
```diff
@@ -337,6 +337,7 @@ public long parseXML(Reader input, TerminationGuard terminationGuard) throws XML
             throw e;
         } finally {
             tx.close();
+            reader.close();
```
loveleif, line 340: If `tx.close()` fails (and I believe it can under certain circumstances) we won't close the reader now. You can avoid this by using try with resources (`try (final var reader = ...) {`), or using something like `IOUtils.closeAll` (from mono repo).

gem-neo4j: This reader isn't auto closeable which is why I didn't wrap it in the try, it is just freeing up resources (not closing the file, which I made a part of a try with resources already), maybe it is fine?

loveleif: Sounds fine :).

PR #624, opened 2024-05-13
[f9JEXqx3] Clean up some docker deprecated config and java deprecated functions values

gem-neo4j:
Reduce the amount of deprecation warnings we get in logs when building/running apoc tests


loveleif (approved):
Nice to see some love given to apoc.

PR #625, opened 2024-05-13
[DBjU6j7r] Clean up deprecated settings

gem-neo4j:
Talked with Misha who deprecated them, he said "please remove usage of that setting its not doing what people assume it was doing and in general case is harmful"


loveleif (approved):
👍

PR #628, opened 2024-05-20
[xHCYeCNX] Added UrlCheck for files

JoelBergstrand:
Added url check for loading files in APOC.

### DIFF
--- a/common/src/main/java/apoc/util/FileUtils.java
+++ b/common/src/main/java/apoc/util/FileUtils.java
@@ -38,6 +38,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLStreamHandler;
 import java.net.URLStreamHandlerFactory;
@@ -49,12 +50,25 @@ import java.util.Optional;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.neo4j.graphdb.security.URLAccessChecker;
+import org.neo4j.graphdb.security.URLAccessValidationError;
 
 /**
  * @author mh
  * @since 22.05.16
  */
 public class FileUtils {
+
+    public static String getLoadFileUrl(String fileName, URLAccessChecker urlAccessChecker)
+            throws MalformedURLException, URLAccessValidationError {
+        URL url;
+        try {
+            url = URI.create(fileName).toURL();
+        } catch (MalformedURLException | IllegalArgumentException e) {
+            String withFile = "file:///" + fileName;
+            url = URI.create(withFile).toURL();
+        }
+        return urlAccessChecker.checkURL(url).getFile();
+    }
 
     public static String getFileUrl(String fileName) throws MalformedURLException {
         try {
@@ -73,7 +87,7 @@ public class FileUtils {
             Map<String, Object> headers,
             String payload,
             URLAccessChecker urlAccessChecker)
-            throws IOException {
+            throws IOException, URLAccessValidationError, URISyntaxException {
         switch (protocol) {
             case s3:
                 return FileUtils.openS3InputStream(urlAddress);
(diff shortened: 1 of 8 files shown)


common/src/main/java/apoc/util/JsonUtil.java
```diff
@@ -123,7 +125,7 @@ public static Stream<Object> loadJson(
                     ? stream
                     : stream.map((value) ->
                             JsonPath.parse(value, getJsonPathConfig(options)).read(path));
-        } catch (IOException e) {
+        } catch (IOException | URISyntaxException | URLAccessValidationError e) {
             if (!failOnError) {
                 return Stream.of();
             } else {
```
JoelBergstrand, line 131: Should the URI/URL exceptions also fallback on an empty stream or always throw an exception?

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
```diff
@@ -137,6 +142,71 @@ public void testRBACOnDeny() throws IOException {
         }
     }
 
+    @Test
+    public void testRBACOnFileDeny() throws IOException {
```
gem-neo4j, line 146: This isn't going to follow the file path though is it? (because it is http scheme)

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
```diff
@@ -137,6 +142,71 @@ public void testRBACOnDeny() throws IOException {
...
+            Assertions.assertThat(e.getMessage()).contains("URLAccessValidationError");
+            Assertions.assertThat(e.getMessage())
+                    .contains("Cause: LOAD on URL '" + url
+                            + "' is denied for user 'testUser' with roles [PUBLIC, editor, test] restricted to");
+        }
+    }
+
+    @Test
+    public void testRBACOnFileDenyAll() throws IOException {
+        String url = "file:///dir/file.txt";
```
gem-neo4j, line 175: We need to also test without the scheme file, absolute paths etc please :)

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
```diff
@@ -137,6 +142,71 @@ public void testRBACOnDeny() throws IOException {
...
+                "apoc.import.graphml($url, {})",
+                "apoc.import.json($url)",
+                "apoc.import.xml($url)");
+
+        for (String loadableAPOCProc : loadableAPOCProcs) {
+            RuntimeException e = assertThrows(
+                    RuntimeException.class,
+                    () -> testCall(testUserSession, "CALL " + loadableAPOCProc, Map.of("url", url), r -> {}));
+
+            Assertions.assertThat(e.getMessage()).contains("IOException");
```
gem-neo4j, line 193: Probably leave out this "working" part, there are other tests for checking these procedures work with files

common/src/main/java/apoc/util/FileUtils.java
```diff
@@ -49,13 +50,26 @@
...
  * @author mh
  * @since 22.05.16
  */
 public class FileUtils {
 
+    public static String getLoadFileUrl(String fileName, URLAccessChecker urlAccessChecker)
+            throws MalformedURLException, URLAccessValidationError {
+        URL url;
+        try {
+            url = URI.create(fileName).toURL();
```
gem-neo4j, line 65: So with this approach we take a file, turn it into a URL, turn it back to a string, then back to a URL? maybe we could have this combined in the checkAllowedUrlAndPinToIP() function instead?

common/src/main/java/apoc/ApocConfig.java
```diff
@@ -324,6 +324,7 @@ public URL checkAllowedUrlAndPinToIP(String url, URLAccessChecker urlAccessCheck
     public void checkReadAllowed(String url, URLAccessChecker urlAccessChecker) throws IOException {
         if (isFile(url)) {
             isImportFileEnabled();
+            checkAllowedUrlAndPinToIP(url, urlAccessChecker);
         } else {
             checkAllowedUrlAndPinToIP(url, urlAccessChecker);
         }
```
gem-neo4j, line 330:
```suggestion
        if (isFile(url)) {
            isImportFileEnabled();
        }
        checkAllowedUrlAndPinToIP(url, urlAccessChecker);
```

common/src/main/java/apoc/util/JsonUtil.java
gem-neo4j: Nah, throw an exception. Then all these places you've added these two, we may as well just throw Exception instead of specifying 3.

gem-neo4j (changes requested):
Hey! Nice work :) Think you broke a few things though 🙈 You can download the summary of failed tests from the PR to take a look

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
JoelBergstrand: You're right, I'll remove it and add some other ones. Should have waited with tagging you until all the old tests were green at least...

it/src/test/java/apoc/it/core/RBACOnLoadTest.java
JoelBergstrand: I've added more tests now, let me know if more tests are needed

gem-neo4j (approved):
Great! Thanks :D

PR #632, opened 2024-05-29
[eqFmcQxi] Fix bug with apoc.export.cypher.* where temp. rel properties were not cleaned up.  

### LINKED ISSUE #631: Cypher script generated by apoc.export.cypher.all leaves the temporary `UNIQUE IMPORT ID REL` relationship properties
## Guidelines

Please note that GitHub issues are only meant for bug reports/feature requests. If you have questions on how to use APOC, please ask on [the Neo4j Discussion Forum](https://community.neo4j.com/c/neo4j-graph-platform/procedures-apoc) instead of creating an issue here.

When executing the cypher script, I expect the temporary `UNIQUE IMPORT ID REL` relationship properties to be removed, as is done for the similar `UNIQUE IMPORT ID` node properties.

All created relationships have an unwanted `UNIQUE IMPORT ID REL` property.

### Simple Dataset (where it's possibile)
This tiny dataset is sufficient:
```
CREATE (n:DummyNode)-[r:DUMMYREL]->(m:OtherDummy)
```

### Steps (Mandatory)

1. Select a database with some data (the exact data is unimportant).

1. Export as cypher, with the `multipleRelationshipsWithType` option set to true:
```
CALL apoc.export.cypher.all(null, {format: 'cypher-shell', multipleRelationshipsWithType: true})
``` 

1. Check the resulting cypher statements. Example:
```
:begin
CREATE CONSTRAINT UNIQUE_IMPORT_NAME FOR (node:`UNIQUE IMPORT LABEL`) REQUIRE (node.`UNIQUE IMPORT ID`) IS UNIQUE;
:commit
CALL db.awaitIndexes(300);
:begin
UNWIND [{_id:1, properties:{}}] AS row
CREATE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id}) SET n += row.properties SET n:OtherDummy;
UNWIND [{_id:0, properties:{}}] AS row
CREATE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id}) SET n += row.properties SET n:DummyNode;
:commit
:begin
UNWIND [{start: {_id:0}, id: 0, end: {_id:1}, properties:{}}] AS row
MATCH (start:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row.start._id})
MATCH (end:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row.end._id})
CREATE (start)-[r:DUMMYREL{`UNIQUE IMPORT ID REL`:row.id}]->(end) SET r += row.properties;
:commit
:begin
MATCH (n:`UNIQUE IMPORT LABEL`)  WITH n LIMIT 20000 REMOVE n:`UNIQUE IMPORT LABEL` REMOVE n.`UNIQUE IMPORT ID`;
:commit
:begin
DROP CONSTRAINT UNIQUE_IMPORT_NAME;
:commit
```

1. Note that the `UNIQUE IMPORT ID` node properties are removed at the end, but not the `UNIQUE IMPORT ID REL` relationship properties.

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: MacOS (with neo4j running in Docker)
  - Neo4j: 5.18.1
  - Neo4j-Apoc: 5.18.0

Lojjs:
Fixes #631 

The bug was only affecting users who have set the `multipleRelationshipsWithType` configuration option to true.


common/src/main/java/apoc/export/cypher/formatter/CypherFormatterUtils.java
```diff
@@ -157,6 +159,18 @@ public static boolean isUniqueLabelFound(Node node, Map<String, Set<String>> uni
...
+     * Returns true if there exists no other relationship with the same start node, end node and type
+     */
+    public static boolean isUniqueRelationship(Relationship rel) {
+        return StreamSupport.stream(
+                        rel.getStartNode()
+                                .getRelationships(Direction.OUTGOING, rel.getType())
+                                .spliterator(),
+                        false)
+                .noneMatch(r -> !r.equals(rel) && r.getEndNode().equals(rel.getEndNode()));
+    }
```
loveleif, line 172: Does this really do what it says it does?

I'm thinking about this situation:

```cypher
create (a), (b)
with a,b 
create (a)-[:R]->(b)
with a,b
create (a)-[:R]->(b)
```

Relationship.equals looks like it's implemented based on id:

```java
@Override
public boolean equals(Object o) {
    return o instanceof Relationship && this.getId() == ((Relationship) o).getId();
}
```

loveleif: Ahh, never mind, now I see it.

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -387,21 +413,35 @@ private void gatherUniqueConstraints() {
         }
     }
 
-    private long countArtificialUniques(Node node) {
+    private long countArtificialUniqueNodes(Node node) {
+        long artificialUniques = 0;
+        artificialUniques = getArtificialUniqueNodes(node, artificialUniques);
+        return artificialUniques;
```
loveleif, line 419:
```suggestion
        return getArtificialUniqueNodes(node, 0);
```

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -387,21 +413,35 @@ private void gatherUniqueConstraints() {
...
+    }
+
+    private long countArtificialUniqueRels(Relationship rel) {
```
loveleif, line 422: This method can also be a single return.

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -414,4 +454,11 @@ private long getArtificialUniques(Node node, long artificialUniques) {
         }
         return artificialUniques;
     }
+
+    private long getArtificialUniqueRels(Relationship rel, long artificialUniques) {
+        if (!CypherFormatterUtils.isUniqueRelationship(rel)) {
+            artificialUniques++;
+        }
```
loveleif, line 461: If it's not unique we increase number of uniques?

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
```diff
@@ -102,7 +104,14 @@ public void export(ExportConfig config, Reporter reporter, ExportFileManager cyp
...
                 break;
             default:
-                artificialUniques += countArtificialUniques(graph.getNodes());
+                artificialUniqueNodes += countArtificialUniqueNodes(graph.getNodes());
+                // In this code path an artificial relationship property is added for each relationship
+                if (exportConfig.isMultipleRelationshipsWithType()) {
+                    artificialUniqueRels += StreamSupport.stream(
+                                    graph.getRelationships().spliterator(), false)
+                            .toList()
+                            .size();
```
loveleif, line 113: No need to materialize the relationships if we only want to count them. 

```suggestion
                    artificialUniqueRels += StreamSupport.stream(
                                    graph.getRelationships().spliterator(), false)
                            .count();
```

common/src/main/resources/exportAll.cypher
```diff
@@ -27,75 +27,75 @@ UNWIND [{_id:160, properties:{born:1967, name:"Julia Roberts", id:"a81fe3a2-68a0
...
..._id:0}, properties:{roles:["Emil"]}}, {start: {_id:38}, id: 5, end: {_id:1}, properties:{roles:["Neo"]}}, {start: {_id:39}, id: 6, end: {_id:1}, properties:{roles:["Trinity"]}}, {start: {_id:40}, id: 7, end: {_id:1}, properties:{roles:["Morpheus"]}}, {start: {_id:41}, id: 8, end: {_id:1}, properties:{roles:["Agent Smith"]}}, {start: {_id:38}, id: 9, end: {_id:2}, properties:{roles:["Neo"]}}, {start: {_id:39}, id: 10, end: {_id:2}, properties:{roles:["Trinity"]}}, {start: {_id:40}, id: 11, end: {_id:2}, properties:{roles:["Morpheus"]}}, {start: {_id:41}, id: 12, end: {_id:2}, properties:{roles:["Agent Smith"]}}, {start: {_id:38}, id: 13, end: {_id:3}, properties:{roles:["Kevin Lomax"]}}, {start: {_id:46}, id: 14, end: {_id:3}, properties:{roles:["Mary Ann Lomax"]}}, {start: {_id:47}, id: 15, end: {_id:3}, properties:{roles:["John Milton"]}}, {start: {_id:49}, id: 16, end: {_id:4}, properties:{roles:["Lt. Daniel Kaffee"]}}, {start: {_id:50}, id: 17, end: {_id:4}, properties:{roles:["Col. Nathan R. Jessup"]}}, {start: {_id:51}, id: 18, end: {_id:4}, properties:{roles:["Lt. Cdr. JoAnne Galloway"]}}, {start: {_id:52}, id: 19, end: {_id:4}, properties:{roles:["Capt. Jack Ross"]}}] AS row
```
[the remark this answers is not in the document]
loveleif: Is it that the bug was that `UNIQUE IMPORT ID REL` properties are supposed to be removed?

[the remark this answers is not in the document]
Lojjs: Yes, I also got confused by this. I think to get the roundtrip to work despite the bug someone (@gem-neo4j ?) had coded this file as if the user already had `UNIQUE IMPORT ID REL` properties in their data before calling APOC (like if they had added it manually). One could argue that in this unlikely event the code is broken after the bug fix as we throw away their properties. However it is the same for nodes and there is a setting to turn off this behaviour, so I would say it is OK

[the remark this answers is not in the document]
Lojjs: Yes, probably. I can try to move it

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
Lojjs: Yes this is what I was confused about in our Slack channel here: https://neo4j.slack.com/archives/C032XMUBAHG/p1716906509842369. The Cypher query we build up will add the `UNIQUE IMPORT ID REL` to all relationships which are not already unique, and we want to calculate how many that will be (i.e. how many we make unique artificially), thus the inversion. This info is then used to know how many batches of deletion we need. Does that make sense? I can add a code comment trying to explain this as we both got confused.

common/src/main/resources/exportAll.cypher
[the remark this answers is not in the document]
Lojjs: It looks like it is potentially used from extended so I don't think it is worth the risk to move it

core/src/main/java/apoc/export/cypher/MultiStatementCypherSubGraphExporter.java
loveleif: Still a bit confusing, but if you feel confident I'm happy :).

loveleif (approved):
Great work!

PR #635, opened 2024-06-17
[VUHeZVxs] Make nodes/rels csv behaviour consistent

gem-neo4j:
In https://github.com/neo4j/apoc/pull/572 the behaviour for backslashes was fixed for nodes, but not for rels. This brings over the fix to relationships as well.


loveleif (approved):
Great! Have you looked for other uses of `CSVReaderBuilder`?

gem-neo4j:
It was only here :)

PR #637, opened 2024-06-25
[hxmGFWvF] Only create a TriggerHandler if triggers are enabled.

Lojjs:
Some of the TriggerHandler code, in particular the start() method was run even if triggers were disabled. This could cause issues in some edge cases as well as impact performance.

Also improve testing of disabled triggers.


gem-neo4j (approved):
Nice work :D

PR #645, opened 2024-07-18
[vMVMuOO7] Export CSV updates

gem-neo4j:
Update behaviour for how ifNeeded works as well as add new config for differentiating null values.

IfNeeded will now correctly quote strings e.g if a string has a quote in it, it will be quoted and escaped.
New config option: differentiateNullls which allows a user to have their null values differentiated from empty strings

Docs: https://github.com/neo4j/docs-apoc/pull/308

### DIFF
--- a/core/src/main/java/apoc/export/csv/CustomCSVWriter.java
+++ b/core/src/main/java/apoc/export/csv/CustomCSVWriter.java
@@ -1,0 +1,34 @@
+package apoc.export.csv;
+
+import com.opencsv.CSVWriter;
+import java.io.Writer;
+
+public class CustomCSVWriter extends CSVWriter {
+
+    private boolean shouldDifferentiateNulls;
+
+    public CustomCSVWriter(
+            Writer writer,
+            char separator,
+            char quoteChar,
+            char escapeChar,
+            String lineEnd,
+            boolean shouldDifferentiateNulls) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/export/csv/CustomCSVWriter.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * We have overridden the openCSV writer so that empty strings are also counted as special.
+     * This is so quotes will be added to them and not to null values if shouldDifferentiateNulls is true.
+     */
+    @Override
+    protected boolean stringContainsSpecialCharacters(String line) {
+        return line.indexOf(this.quotechar) != -1
+                || line.indexOf(this.escapechar) != -1
```
anderson4j, line 28: Think both approaches work, but why not contains here too?

anderson4j (approved):
Had a small comment that doesnt matter much (but if there's an interesting answer do lmk!). Good stuff!

core/src/main/java/apoc/export/csv/CustomCSVWriter.java
gem-neo4j: This is a direct copy from the actual CSV writer that it overrides, plus my additions, I figured if I left it looking the same as the original, then any future updates are easier to make if needed :)

PR #648, opened 2024-07-29
[FiO72wgR] Remove old github actions

gem-neo4j:
APOC is now connected to TC, so this removes the testing part of the github actions


loveleif (approved):
👍

PR #649, opened 2024-07-30
Removing references to legacy builds in TeamCity

omar952:


gem-neo4j (approved):
LGTM, the failing test is actually broken, but not by you :) I will fix separately so feel free to ignore it :)

gem-neo4j (commented):
Looks good, we also update the version in the README :)

Issue #651, opened 2024-08-08
apoc.convert.toTree marked as deprecated with alternative not in documentation


Nio181:
## Expected Behavior
For Cypher READ queries I use the procedure "apoc.convert.toTree" to get the result in a JSON-format. As this procedure is mentioned in the documentation, I do not expect errors. <https://neo4j.com/docs/apoc/current/overview/apoc.convert/apoc.convert.toTree/>

## Actual Behavior
I get the result as expected, but also the warning:
`The query used a deprecated procedure. ('apoc.convert.toTree' has been replaced by 'apoc.paths.toJsonTree')`
Replacing the procedure with `apoc.paths.toJsonTree` does not lead to an error, but does not return a result either. In addition to that, in the documentation the procedure `apoc.paths.toJsonTree` is not listed.

## How to Reproduce the Problem
I use the version neo4-community-5.22.0 running locally on my machine with the Browser UI. The command that leads to this problem is 
`MATCH path=(n:Actor {ID: '4c8c2bbd-ae5e-478f-81e5-494c063cb645'})
WITH COLLECT(path) AS paths
CALL apoc.convert.toTree(paths)
YIELD value RETURN value`

### Simple Dataset
For performing the same command as I did, you just need the following Cypher statement
```
CREATE(n:Actor {ID: '4c8c2bbd-ae5e-478f-81e5-494c063cb645'})
```

## Screenshots
![image](https://github.com/user-attachments/assets/49a98159-2e83-4d94-afc3-9b599a61f707)

### Versions

  - OS: MacOS Sonoma 14.6.1
  - Neo4j: neo4j-community-5.22.0
  - Neo4j-Apoc: apoc-5.21.0

gem-neo4j:
Hey!

Thanks for writing in, this replacement procedure was very recently added and it appears the documentation was never published for it, sorry about that, we have published it [now](https://neo4j.com/docs/apoc/current/overview/apoc.paths/apoc.paths.toJsonTree/) :) As for the Path of size 1 not returning results, that seems to be an oversight, I have put up a PR for a fix to this, it will hopefully be released in the 5.24 APOC release.

gem-neo4j:
The fix has been merged, and will be in the 5.24 release (and subsequent 4.4 release as well) :)

PR #653, opened 2024-08-12
[NOID] Return a node even if no relationships exist

gem-neo4j:
Return the root node as well as a single node is a path of 1.

Fixes issue: https://github.com/neo4j/apoc/issues/651


JoelBergstrand (approved):
LGTM

PR #665, opened 2024-10-08
[d4rLFQqM] Align all export modes with quoting consistency

gem-neo4j:
When bulkImport was set to true, the quoting option the user input no longer worked. This PR makes the quoting work again and makes sure the defaults align with how it works for streaming and when bulkImport is false.

### DIFF
--- a/core/src/main/java/apoc/export/csv/ExportCSV.java
+++ b/core/src/main/java/apoc/export/csv/ExportCSV.java
@@ -84,10 +84,11 @@ public class ExportCSV {
                     {
                             stream = false :: BOOLEAN,
                             batchSize = 20000 :: INTEGER,
-                            bulkImport = true :: BOOLEAN,
+                            bulkImport = false :: BOOLEAN,
                             timeoutSeconds = 100 :: INTEGER,
                             compression = 'None' :: STRING,
                             charset = 'UTF_8' :: STRING,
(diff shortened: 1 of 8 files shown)


core/src/main/java/apoc/export/csv/ExportCSV.java
```diff
@@ -84,10 +84,11 @@ public Stream<ExportProgressInfo> all(
                     {
                             stream = false :: BOOLEAN,
                             batchSize = 20000 :: INTEGER,
-                            bulkImport = true :: BOOLEAN,
+                            bulkImport = false :: BOOLEAN,
```
gem-neo4j, line 87: Note to reviewer, I didn't change the default, it was this signature that was wrong :)

hvub (approved):
I think that looks fine. 

Somethings noticed in passing [here](https://github.com/neo4j/apoc/pull/665/files#diff-ee25230803c5903338f189c185effd6ebc4612fbbdb5da28d34cb8604934bd52R203-R208):

```
    private void preventBulkImport(ExportConfig config) {
        if (config.isBulkImport()) {
            throw new RuntimeException(
                    "You can use the `bulkImport` only with apoc.export.all and apoc.export.csv.graph");
        }
    }
```

shouldn't the exception message say `apoc.export.csv.all` instead?

PR #666, opened 2024-10-09
[yVM4sJbH]  Dev fix round trip for csv loading

gem-neo4j:
Long and Double types can be exported but not imported, so this PR allows for them to be imported (if CSV header says prop:long or prop:double for example).

This PR builds on top of https://github.com/neo4j/apoc/pull/665

### DIFF
--- a/common/src/main/java/apoc/load/Mapping.java
+++ b/common/src/main/java/apoc/load/Mapping.java
@@ -110,8 +110,10 @@ public class Mapping {
                 return DateValue.parse(value).asObjectCopy();
             case DURATION:
                 return DurationValue.parse(value);
+            case LONG:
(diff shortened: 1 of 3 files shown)


hvub (approved):
Looks good to me.

A side note: I am still surprised that `apoc.export` procedures have a `bulkImport` option, but

1. that is not subject of this PR
2. there may be good reason for it, which I have not investigated
3. we may not want to change it no matter what

gem-neo4j:
@hvub hehe be prepared for more surprises the deeper you dive into APOC 😆 Anyway, afaik the bulkImport config means that it prepares files in such a way that the user can use them in a bulkImport later, not that it exports them in bulk :)

PR #667, opened 2024-10-10
[v8imq7eo] Fix coll array handling

gem-neo4j:
Internally neo uses different structures when storing lists than Cypher may use, so we need to make sure we convert any object comparisons appropriately in the coll functions


loveleif:
Since Gem is out anyway I'm waiting with the review until next week unless someone wants this asap.

loveleif (approved):
Sweet

PR #671, opened 2024-10-21
[NOID] Update testContainers

gem-neo4j:
In the APOC Extended updates, I updated this to the latest version as 1.19.7 has a bug, better to update core directly over that bug to make sure we don't run into the same issues as extended!


loveleif (approved):
Great! I'm a bit worried about the versions moving around, some that I recognize from the mono repo, I hope there's not a risk of shadowing something.

PR #674, opened 2024-10-23
[bkym7TcZ] Optimise apoc.refactor.cloneSubgraph

loveleif:
Add option `createNodesInNewTransactions` to `apoc.refactor.cloneSubgraph` that let's you control if nodes are created in inner transactions. Previous behavior was to create each node in a new transaction (relationships were not though, surprisingly). New default behavior is to use the outer transaction for all writes (`createNodesInNewTransactions = true` gives the old behavior still).

### DIFF
--- a/core/src/test/java/apoc/refactor/GraphRefactoringTest.java
+++ b/core/src/test/java/apoc/refactor/GraphRefactoringTest.java
@@ -24,13 +24,11 @@ import static apoc.util.TestUtil.testCallCount;
 import static apoc.util.TestUtil.testCallEmpty;
 import static apoc.util.TestUtil.testResult;
 import static apoc.util.Util.isSelfRel;
-import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonList;
-import static org.hamcrest.CoreMatchers.equalTo;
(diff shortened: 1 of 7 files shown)


core/src/test/java/apoc/refactor/GraphRefactoringTest.java
```diff
@@ -333,27 +330,31 @@ public void deleteAndReconnectWithMergeRelConfigAndPropertiesCombine() {
...
-                    assertEquals("f", rel1.getProperty("e"));
-                    assertThat(Arrays.asList((String[]) rel1.getProperty("a")), containsInAnyOrder("b", "d"));
-                    assertEquals("h", rel1.getProperty("g"));
-                    Relationship rel2 = rels.get(1);
-                    assertEquals("GAMMA_DELTA", rel2.getType().name());
-                    assertThat(Arrays.asList((String[]) rel2.getProperty("aa")), containsInAnyOrder("one", "bb"));
-                    assertEquals("ff", rel2.getProperty("ee"));
-                    assertEquals("dd", rel2.getProperty("cc"));
-                    assertNotNull(row.get("nodes"));
+                    assertThat(row.get("nodes"))
```
loveleif, line 333: I had some test failures of the unhelpful kind "false was not equal to true". So refactored some tests to help me out unserstanding that.

JoelBergstrand (approved):
Lgtm!

PR #675, opened 2024-10-28
[3Hl7fTsQ]  Dev fix misleading name column

gem-neo4j:
The name column was different between these 2 equivalent procedures, and wasn't even actually giving the name of the constraint/index. So this makes it more consistent with SHOW CONSTRAINTS/INDEXES


loveleif (approved):
👍

gem-neo4j:
@loveleif in order to fix a failing test I had to make quite a few changes, when you have time can you review the second commit? :)

PR #682, opened 2024-11-04
[EUtjTRHb] Better error handling and logging during cleanup of triggers

loveleif:
- Add better logging when apoc fails to cleanup triggers during database drop.
- Ignore trigger cleanup when system db is unavailable.

### DIFF
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -189,8 +189,9 @@ public class CypherInitializer implements AvailabilityListener {
                     tx.commit();
                 } catch (Exception e) {
                     userLog.error(
-                            "Failed to cleanup apoc triggers during database drop of %s: %s".formatted(dbName, e), e);
-                    throw e;
+                            "Failed to cleanup apoc triggers during database drop of %s, please run `apoc.trigger.dropAll` manually to cleanup: %s"
+                                    .formatted(dbName, e),
+                            e);
                 }
             }
         }
--- a/core/src/main/java/apoc/cypher/CypherInitializer.java
+++ b/core/src/main/java/apoc/cypher/CypherInitializer.java
@@ -29,15 +29,14 @@ import apoc.util.collection.Iterators;
 import apoc.version.Version;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -174,11 +173,26 @@ private class SystemFunctionalityListener implements DatabaseEventListener {
...
-                        .forEachRemaining(Node::delete);
-            });
+            final var dbName = eventContext.getDatabaseName();
+            if (!Objects.equals(dbName, db.databaseName())) {
+                if (!db.isAvailable()) {
+                    userLog.warn(
+                            "Database %s not available, skipping apoc trigger cleanup of database %s.",
+                            db.databaseName(), dbName);
+                    return;
+                }
```
loveleif, line 183: We risk leaving triggers laying around for dropped databases. But what else can we do if system db is unavailable?

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -174,11 +173,26 @@ private class SystemFunctionalityListener implements DatabaseEventListener {
...
+                try (final var tx = db.beginTx()) {
+                    for (Label label : SystemLabels.values()) {
+                        tx.findNodes(label, database.name(), eventContext.getDatabaseName())
+                                .forEachRemaining(Node::delete);
+                    }
+                    tx.commit();
+                } catch (Exception e) {
+                    userLog.error(
+                            "Failed to cleanup apoc triggers during database drop of %s: %s".formatted(dbName, e), e);
+                    throw e;
```
loveleif, line 193: Should we even swallow this exception?

gem-neo4j: Yeah, I think swallow it

gem-neo4j (approved):
LGTM, I think swallowing would be fine (especially because triggers don't exist in Aura)

core/src/main/java/apoc/cypher/CypherInitializer.java
loveleif: Makes sense, updated the PR to ignore this error.

PR #685, opened 2024-11-13
[NOID] show the :: without an irritating space in between

hvub:
The README shows `name : : TYPE` with an irritating space in between.

Get rid of the space and make the README show `name :: TYPE`.


gem-neo4j (approved):
🙌

PR #688, opened 2024-11-15
APOC to jdk 21

gem-neo4j:
Neo has moved to Java 21, so we will follow suit!

Gradlew 8.5 is needed to get Java 21 full support.
There were some bugs in spotless so need to specify more specifically which versions to use.

Also need to version bump, 2024.12.0


gem-neo4j:
[Move to Java 21?](https://trello.com/c/W0mpsXW7/2559-move-to-java-21)

Lojjs (approved):

PR #696, opened 2024-12-03
Migrate all common over to core to remove dependencies between core and extended

gem-neo4j:
This PR is moving all the files under common into core, another PR will be made for fixing up issues in extended. The goal is to not need to depend on common as a module. This means we can change things a lot more freely in core without breaking Extended. It should also fix some path issues people have had with modules in core/common/extended.


gem-neo4j:
[Duplicate Common into Extended to remove dependency between the two repos](https://trello.com/c/A2Ltc4Ai/2517-duplicate-common-into-extended-to-remove-dependency-between-the-two-repos)

PR #709, opened 2024-12-19
Fix config maps for graphml procs

gem-neo4j:
The config values were incorrect; [docs](https://github.com/neo4j/docs-apoc/pull/348)


hvub (approved):
🎉🎉

PR #710, opened 2024-12-23
Add other types to give a more robust grouping proc

gem-neo4j:
https://github.com/neo4j/apoc/issues/116

### DIFF
--- a/core/src/main/java/apoc/nodes/Grouping.java
+++ b/core/src/main/java/apoc/nodes/Grouping.java
@@ -21,16 +21,23 @@ package apoc.nodes;
 import static java.util.Collections.*;
 
 import apoc.Pools;
+import apoc.convert.ConvertUtils;
 import apoc.result.VirtualNode;
 import apoc.result.VirtualRelationship;
 import apoc.util.Util;
 import apoc.util.collection.Iterables;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetTime;
+import java.time.ZonedDateTime;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.neo4j.exceptions.ArithmeticException;
 import org.neo4j.graphdb.*;
 import org.neo4j.logging.Log;
 import org.neo4j.procedure.Context;
@@ -38,6 +45,8 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.NotThreadSafe;
 import org.neo4j.procedure.Procedure;
+import org.neo4j.values.storable.DurationValue;
+import org.neo4j.values.storable.PointValue;
 
 /**
  * @author mh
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/nodes/Grouping.java
```diff
@@ -368,6 +377,10 @@ private <C extends Collection<T>, T extends Entity> C fixAggregates(C pcs) {
                     double[] values = (double[]) v;
                     entry.setValue(values[1] == 0 ? 0 : values[0] / values[1]);
                 }
+                if (k.matches("^avg_.+") && v instanceof DurationValue) {
```
loveleif, line 380: Regexing the property name, not very rubust, but I guess this is the way of the apoc.

core/src/main/java/apoc/nodes/Grouping.java
```diff
@@ -368,6 +377,10 @@ private <C extends Collection<T>, T extends Entity> C fixAggregates(C pcs) {
...
+                    Long count = ((Number) pc.getProperty(k + "_count", 0)).longValue();
+                    entry.setValue(divDurationValue((DurationValue) v, count));
+                }
```
loveleif, line 383:
```suggestion
                if (k.matches("^avg_.+") && v instanceof DurationValue duration) {
                    Long count = ((Number) pc.getProperty(k + "_count", 0)).longValue();
                    entry.setValue(divDurationValue(duration, count));
                }
```

core/src/main/java/apoc/nodes/Grouping.java
```diff
@@ -368,6 +377,10 @@ private <C extends Collection<T>, T extends Entity> C fixAggregates(C pcs) {
                     double[] values = (double[]) v;
                     entry.setValue(values[1] == 0 ? 0 : values[0] / values[1]);
                 }
+                if (k.matches("^avg_.+") && v instanceof DurationValue) {
+                    Long count = ((Number) pc.getProperty(k + "_count", 0)).longValue();
```
loveleif, line 381: Is it intentional to use `.longValue()` here? I thinking about `Double`s for example, that will get rounded by that call.

core/src/main/java/apoc/nodes/Grouping.java
```diff
@@ -395,28 +423,36 @@ private void aggregate(Entity pc, Map<String, List<String>> aggregations, Map<St
                                 pc.setProperty(key, ((Number) pc.getProperty(key, 0)).longValue() + 1);
                                 break;
                             case "sum":
-                                pc.setProperty(
-                                        key, ((Number) pc.getProperty(key, 0)).doubleValue() + Util.toDouble(value));
+                                if (value instanceof DurationValue) {
+                                    DurationValue dv =
+                                            (DurationValue) pc.getProperty(key, DurationValue.duration(0, 0, 0, 0));
+                                    pc.setProperty(key, ((DurationValue) value).add(dv));
```
loveleif, line 429:
```suggestion
                                if (value instanceof DurationValue duration) {
                                    DurationValue dv =
                                            (DurationValue) pc.getProperty(key, DurationValue.duration(0, 0, 0, 0));
                                    pc.setProperty(key, (duration.add(dv));
```

core/src/main/java/apoc/nodes/Grouping.java
```diff
@@ -426,6 +462,145 @@ private void aggregate(Entity pc, Map<String, List<String>> aggregations, Map<St
...
+                || (prop instanceof DurationValue && value instanceof DurationValue)
+                || (prop instanceof String && value instanceof String)
+                || (prop instanceof Boolean && value instanceof Boolean)
+                || (prop instanceof Number && value instanceof Number)
+                || ((prop instanceof Collection || prop.getClass().isArray())
+                        && (value instanceof Collection || value.getClass().isArray()))
+                || (prop instanceof PointValue && value instanceof PointValue);
+    }
+
+    private boolean compareValues(Object prop, Object value) {
```
loveleif, line 508: Could use `org.neo4j.values.storable.Values.COMPARATOR` here to not have to maintain this in apoc?

```java
return switch (Values.COMPARATOR.ternaryCompare(Value.of(a), Value.of(b))) {
  case UNDEFINED -> ...;
  case EQUAL -> ...;
  case GREATER_THAN -> ...;
  case SMALLER_THAN -> ...;
}
```

loveleif (approved):
👍

core/src/main/java/apoc/nodes/Grouping.java
loveleif: Compiling a regex pattern is a bit expensive though (I don't think they're cached in java) so we could consider replacing this with `k.startsWith("avg_") && k.size() > "avg_".length`. Or to compile the Pattern once in a field and reuse.

core/src/main/java/apoc/nodes/Grouping.java
gem-neo4j: I did not know this existed, looks interesting thanks! I'll play with it :D

PR #712, opened 2025-01-02
Add option to wrap a nodes id to the apoc.create.virtual.fromNode

gem-neo4j:
Apparently the original idea of this function was to just wrap real nodes aka keep their ids, to not break any existing behaviour I have added that as a config option instead

docs: https://github.com/neo4j/docs-apoc/pull/363


JoelBergstrand (approved):
Lgtm

PR #719, opened 2025-01-08
Make use of new database setting for determining conf directory

JoelBergstrand:


common/src/main/java/apoc/ApocConfig.java
```diff
@@ -188,30 +188,7 @@ public void init() {
...
-                    .findFirst()
-                    .orElse(defaultPath);
-            if (defaultPath.equals(neo4jConfFolder)) {
-                log.info("cannot determine conf folder from sys property %s, assuming %s", command, defaultPath);
-            } else {
-                log.info("from system properties: NEO4J_CONF=%s", neo4jConfFolder);
-            }
-            return neo4jConfFolder;
-        }
+        return neo4jConfig.get(configuration_directory).toString();
```
gem-neo4j, line 191: I can't comment directly on the correct line, but where this is used now has an incorrect comment, can you fix that up? :) also the log info is wrong now

JoelBergstrand: Could the log say something like this?
```suggestion
        String neo4jConfFolder = neo4jConfig.get(configuration_directory).toString();
        log.info("from database setting: server.directories.configuration=%s", neo4jConfFolder);
        return neo4jConfFolder;
```

gem-neo4j: Hmm actually, that line is about setting a system property. Do we even need that?

JoelBergstrand: I don't think we need the loging any more. Omitting it and updating the comment.

loveleif (approved):
Super

PR #722, opened 2025-01-27
Remove resolvers on db shutdown

gem-neo4j:
On shutdown, remove the resolvers which hold memory references


loveleif (approved):
I agree, this looked suspicious!

loveleif (commented):
It would be nice with a test that:

1. Creates a db.
2. Install triggers on that db (and perhaps other things that touches RegisterComponentFactory/ApocExtensionFactory).
3. Stop database.
4. Start database.
5. Make sure the trigger still fires.

PR #725, opened 2025-01-29
Add ability to test different cypher versions globally

gem-neo4j:
Add config and env variable to be able to test Cypher Versions - ready to fix all Cypher 25 test failures

### DIFF
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -111,6 +111,7 @@ import org.neo4j.graphdb.schema.IndexType;
 import org.neo4j.graphdb.security.URLAccessChecker;
 import org.neo4j.graphdb.security.URLAccessValidationError;
 import org.neo4j.internal.schema.ConstraintDescriptor;
+import org.neo4j.kernel.api.QueryLanguage;
 import org.neo4j.kernel.impl.coreapi.InternalTransaction;
 import org.neo4j.kernel.impl.util.ValueUtils;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
@@ -1360,5 +1361,12 @@ public class Util {
             }
         }
         return result;
+    }
+
+    // Get the current supported query language versions, if this list changes
+    // this function will error, on error please update!
+    public static List<String> getCypherVersions() {
+        assert QueryLanguage.ALL.size() == 2;
+        return List.of("5", "25");
     }
 }
--- a/common/src/main/java/apoc/util/Util.java
+++ b/common/src/main/java/apoc/util/Util.java
@@ -1366,7 +1366,11 @@ public class Util {
     // Get the current supported query language versions, if this list changes
     // this function will error, on error please update!
     public static List<String> getCypherVersions() {
-        assert QueryLanguage.ALL.size() == 2;
-        return List.of("5", "25");
+        return QueryLanguage.ALL.stream()
+                .map(l -> switch (l) {
+                    case QueryLanguage.CYPHER_5 -> "5";
+                    case QueryLanguage.CYPHER_25 -> "25";
+                })
+                .toList();
     }
 }
--- a/core/src/test/java/apoc/ArgumentDescriptionsTest.java
+++ b/core/src/test/java/apoc/ArgumentDescriptionsTest.java
@@ -106,7 +106,6 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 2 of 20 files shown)


test-utils/src/main/java/org/neo4j/test/rule/DbmsRule.java
```diff
@@ -153,6 +154,19 @@ protected void after() {
...
     private void create() {
         databaseBuilder = newFactory();
+
+        // Allow experimental versions of Cypher and set Cypher Default Version
+        globalConfig.put(GraphDatabaseInternalSettings.enable_experimental_cypher_versions, true);
+
+        // A test may set this for the entire file itself, so we shouldn't override that
+        if (!globalConfig.containsKey(GraphDatabaseInternalSettings.default_cypher_version)) {
+            String cypherVersionEnv = System.getenv()
+                    .getOrDefault("CYPHER_VERSION", GraphDatabaseInternalSettings.CypherVersion.Cypher25.name());
```
gem-neo4j, line 164: Should we leave this as 25 going forward?

The plan is to ask for the TC builds to use the env to test both Cypher 5 and 25 in PR builds :)

common/src/main/java/apoc/util/Util.java
```diff
@@ -1361,4 +1362,11 @@ public static <T> T withBackOffRetries(Supplier<T> func, long initialTimeout, lo
         }
         return result;
     }
+
+    // Get the current supported query language versions, if this list changes
+    // this function will error, on error please update!
+    public static List<String> getCypherVersions() {
+        assert QueryLanguage.ALL.size() == 2;
+        return List.of("5", "25");
+    }
```
loveleif, line 1371: Could also be a compilation failure:

```suggestion
    public static List<String> getCypherVersions() {
        return QueryLanguage.ALL.stream()
          .map(l -> switch(l) {
            case Cypher5 -> "5";
            case Cypher25 -> "25";
          })
          .toList();
    }
```

core/src/test/java/apoc/ArgumentDescriptionsTest.java
```diff
@@ -202,9 +200,11 @@ public void teardown() {
 
     @Test
     public void functionArgumentDescriptionsDefaultVersion() throws IOException {
+        // TODO: When Cypher command for getting default is available use that here
+        // for now we just force this to use the preset default
         assertResultAsJsonEquals(
                 CypherVersion.Default,
-                showFunctions(null),
+                showFunctions(CypherVersion.Default),
```
loveleif, line 207: `CypherVersion.Default` is being removed. If you really need a default one you can use:

```java
switch (GraphDatabaseSettings.default_langage) match {
  case Cypher5 => CypherVerision.Cypher5
  case Cypher25 => CypherVerision.Cypher25
}
```

core/src/test/java/apoc/ArgumentDescriptionsTest.java
```diff
@@ -223,9 +223,11 @@ public void functionArgumentDescriptions() throws IOException {
 
     @Test
     public void procedureArgumentDescriptionsDefaultVersion() throws IOException {
+        // TODO: When Cypher command for getting default is available use that here
+        // for now we just force this to use the preset default
         assertResultAsJsonEquals(
                 CypherVersion.Default,
-                showProcedures(null),
+                showProcedures(CypherVersion.Default),
```
loveleif, line 230: Same here

core/src/test/java/apoc/export/arrow/ArrowTest.java
```diff
@@ -60,7 +64,10 @@ public class ArrowTest {
     public static DbmsRule db = new ImpermanentDbmsRule()
             .withSetting(
                     GraphDatabaseSettings.load_csv_file_url_root,
-                    directory.toPath().toAbsolutePath());
+                    directory.toPath().toAbsolutePath())
+            .withSetting(
+                    GraphDatabaseInternalSettings.default_cypher_version,
```
loveleif, line 69: This setting is being replaced by `GraphDatabaseSettings.default_language`. Unfortunately that is not hooked up yet ([PR](https://github.com/neo-technology/neo4j/pull/29353)), so go ahead and merge this and I sort it out later.

test-utils/src/main/java/org/neo4j/test/rule/DbmsRule.java
loveleif: I vote for using 5 as default for now and setting it to 25 when mono repo has flipped.

loveleif (approved):
Nice!

PR #732, opened 2025-02-04
testTerminateCommit: stop test going out of memory

gem-neo4j:
This test was running continuously 1000 creates, using a MERGE instead it hopefully won't run out of memory as often


loveleif (approved):
Great! I've used apoc sleep for this purpose in the past, but nothing wrong with this.

PR #733, opened 2025-02-05
Add close() method to `LimitedSizeInputStream`

hvub:
Fixes "graphml import query is locking the graphml file after execution"

`LimitedSizeInputStream` misses a close method causing file handles on imports to remain open (until collected by GC).

This PR adds the missing close method, which fixes the problem.


loveleif (approved):
Sweet! Do we need to backport?

Issue #721, opened 2025-01-26
Multiple apoc.create.virtual.fromNode with the same node have the same id


abccbaandy:
## Guidelines

Please note that GitHub issues are only meant for bug reports/feature requests. If you have questions on how to use APOC, please ask on [the Neo4j Discussion Forum](https://community.neo4j.com/c/neo4j-graph-platform/procedures-apoc) instead of creating an issue here.

## Expected Behavior (Mandatory)
All virtual nodes should have different id

## Actual Behavior (Mandatory)
The ids are all the same.
Not just id, the element id is the same too.
Also why there is no `apoc.node.elementId` apoc?

The custom property(myUUID) is work, means it's possible to create two virtual node from the same node, right?

### Simple Dataset (where it's possibile)

```
//Insert here a set of Cypher statements that helps us to reproduce the problem
Just use Movie dataset
```

### Steps (Mandatory)

  1.
````
MATCH (p:Person)--(n:Movie {title:"A Few Good Men"}) 
LIMIT 2
WITH p, collect(n) as movies
CALL(movies) {
UNWIND movies as m
WITH apoc.create.virtual.fromNode(m, keys(m))  as node 
CALL apoc.create.setProperties(node, ["id", "myUUID", "myElementId"], [apoc.node.id(node), randomUUID(), elementId(node)]) 
YIELD node as newNode
RETURN collect(newNode) as newList
} 
RETURN  p, newList
````

## Screenshots (where it's possibile)
````
1
(:Person
  { born: 1962,
    name: "Tom Cruise" }
)
[(:Movie
  { myUUID: "1bfeb31f-4884-432a-827b-e234c6322efc",
    tagline: "In the heart of the nation's capital, in a courthouse of the U.S. government, one man will stop at nothing to keep his honor, and one will stop at nothing to find the truth.",
    id: -16,
    myElementId: "4:bd26d4b2-dbe7-4cf6-993b-fe76f872cb78:15",
    title: "A Few Good Men",
    released: 1992 }
)]
2
(:Person
  { born: 1937,
    name: "Jack Nicholson" }
)
[(:Movie
  { myUUID: "91a8b8a7-8abb-4fdb-a80c-8b4d04c42ac1",
    tagline: "In the heart of the nation's capital, in a courthouse of the U.S. government, one man will stop at nothing to keep his honor, and one will stop at nothing to find the truth.",
    id: -16,
[...]

PR #743, opened 2025-02-24
Fix virtual node id working

gem-neo4j:
Fixes https://github.com/neo4j/apoc/issues/721.

If it isn't specified that the virtual node should keep the original id, the new virtual nodes will get new negative ids (unique)

### DIFF
--- a/common/src/main/java/apoc/result/VirtualNode.java
+++ b/common/src/main/java/apoc/result/VirtualNode.java
@@ -67,9 +67,8 @@ public class VirtualNode implements Node {
         Objects.requireNonNull(node, ERROR_NODE_NULL);
         final long id = node.getId();
         // if node is already virtual, we return the same id
-        this.id = id < 0 || wrapNodeIDs ? id : -id - 1;
+        this.id = id < 0 || wrapNodeIDs ? id : MIN_ID.decrementAndGet();
         // to not overlap this nodes ids with ids from VirtualNode(Label[] labels, Map<String, Object> props)
-        MIN_ID.updateAndGet(x -> Math.min(x, this.id));
         this.labels.addAll(Util.labelStrings(node));
         String[] keys = propertyNames.toArray(new String[propertyNames.size()]);
         this.props.putAll(node.getProperties(keys));
--- a/core/src/test/java/apoc/create/CreateTest.java
+++ b/core/src/test/java/apoc/create/CreateTest.java
@@ -416,6 +416,35 @@ public class CreateTest {
     }
 
     @Test
+    public void test2VirtualFromNodesHaveUniqueIDs() {
+        testCall(
+                db,
+                """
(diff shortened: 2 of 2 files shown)


gem-neo4j:
[apoc.create.fromVirtualNode id bug](https://trello.com/c/gMDy3M9F/2594-apoccreatefromvirtualnode-id-bug)

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -415,6 +415,35 @@ public void testVirtualFromNodeFunction() {
                 });
     }
 
+    @Test
+    public void test2VirtualFromNodesHaveUniqueIDs() {
```
hvub, line 419: Would

```
CREATE (n1:Cat { name:'Maja', born: 2023 } )
WITH n1
UNWIND [1,2,3] AS i
RETURN apoc.create.virtual.fromNode(n1, ['name']) AS node1
```

be a useful test? I would expect it to return three new vnodes with the same property and label but different ids.

Maybe such test already exist. I have looked at the whole class.

common/src/main/java/apoc/result/VirtualNode.java
```diff
@@ -67,9 +67,8 @@ public VirtualNode(Node node, List<String> propertyNames, boolean wrapNodeIDs) {
         Objects.requireNonNull(node, ERROR_NODE_NULL);
         final long id = node.getId();
         // if node is already virtual, we return the same id
-        this.id = id < 0 || wrapNodeIDs ? id : -id - 1;
+        this.id = id < 0 || wrapNodeIDs ? id : MIN_ID.decrementAndGet();
         // to not overlap this nodes ids with ids from VirtualNode(Label[] labels, Map<String, Object> props)
-        MIN_ID.updateAndGet(x -> Math.min(x, this.id));
         this.labels.addAll(Util.labelStrings(node));
         String[] keys = propertyNames.toArray(new String[propertyNames.size()]);
         this.props.putAll(node.getProperties(keys));
```
hvub, line 74: Referring to:

```
this.elementId = node.getElementId();
```

just keeping the element id. Is that right?

It looks like a potential source of confusion, particularly, in a scenario like the one mentioned in my other comment. Is it, at least, worthwhile mentioning this behavior in the documentation?

I know this is outside the fix.

hvub (approved):
This appears to be a straightforward fix. Nice.

I have noted two observations that jumped to mind in passing. I leave it to your discretion to decide what to do with that.

common/src/main/java/apoc/result/VirtualNode.java
gem-neo4j: Yep, I agree, I wanted to keep that separate from this fix though :)

core/src/test/java/apoc/create/CreateTest.java
gem-neo4j: Ah yes, that is useful and was what I was meant to test 😅 not sure why I wrote the test I did

Issue #68, opened 2022-09-01
Procedure `apoc.meta.subgraph` return wrong type with `includeLabels` not existent


neo-technology-build-agent:
<a href="https://github.com/vga91"><img src="https://avatars.githubusercontent.com/u/25103389?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [vga91](https://github.com/vga91)**
_Monday May 17, 2021 at 10:35 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1906_

----

## Expected Behavior (Mandatory)

Should return `nodes: []` and `relationships: []`

## Actual Behavior (Mandatory)

The procedure returns a node with dataset total count.

### Simple Dataset (where it's possibile)

```
CREATE (:Foo), (:Bar)

```

### Steps (Mandatory)

  1. `CALL apoc.meta.subGraph({labels:["Test"],rels:[],excludes:[]})`
  2. Returns a `"Test"` node with count 2, even if the label does not exist 
 

## Screenshots (where it's possibile)
<img width="799" alt="Schermata 2021-05-17 alle 12 33 26" src="https://user-images.githubusercontent.com/25103389/118475070-1719f600-b70c-11eb-9669-01de77746b96.png">

Currently used versions

### Versions

  - Neo4j: 4.0.11
  - Neo4j-Apoc: 4.0.0.18

neo-technology-build-agent:
<a href="https://github.com/vga91"><img src="https://avatars.githubusercontent.com/u/25103389?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [vga91](https://github.com/vga91)**
_Monday May 17, 2021 at 12:57 GMT_

----

Related to #1491 (Same root cause, but without rels)

hvub:
Here is a little more detailed variant (on 5.25.1)

Setup:
```
MATCH (n) DETACH DELETE n
WITH COUNT(1) as foo
CREATE (:Foo), (:Bar), (:Test) // label Test created here
WITH COUNT(1) as foo
MATCH (t:Test) DELETE t // node with label Test delete again
WITH COUNT(1) as foo
MATCH (n)
RETURN n
```

Query 1:
```
CALL apoc.meta.subGraph({includeLabels:["Test"],rels:[],excludes:[]})
```
=> correct result

Query 2:
```
CALL apoc.meta.subGraph({includeLabels:["X"],rels:[],excludes:[]})
```
[...]

PR #744, opened 2025-02-26
Fix apoc.meta.subgraph for non-existing labels and types

gem-neo4j:
A non-existing label or type has the id of -1, which so happens to be the id of the constant for "ANY_LABEL" and "ANY_TYPE" which meant giving non-existing ones in would return the count of the db 😅

Fixes: https://github.com/neo4j/apoc/issues/68

### DIFF
--- a/common/src/main/java/org/neo4j/cypher/export/DatabaseSubGraph.java
+++ b/common/src/main/java/org/neo4j/cypher/export/DatabaseSubGraph.java
@@ -142,24 +142,39 @@ class CountOptimisedDatabaseSubGraph extends DatabaseSubGraph {
 
     @Override
     public long countsForNode(Label label) {
-        return read.countsForNode(tokenRead.nodeLabel(label.name()));
+        int nodeLabelID = tokenRead.nodeLabel(label.name());
+        if (nodeLabelID < 0) {
+            return 0;
+        }
+        return read.countsForNode(nodeLabelID);
     }
 
     @Override
     public long countsForRelationship(RelationshipType type, Label end) {
-        return read.countsForRelationship(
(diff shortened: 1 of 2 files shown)


gem-neo4j:
[Procedure apoc.meta.subgraph return wrong type with includeLabels not existent](https://trello.com/c/MJyeWGqg/2570-procedure-apocmetasubgraph-return-wrong-type-with-includelabels-not-existent)

common/src/main/java/org/neo4j/cypher/export/DatabaseSubGraph.java
```diff
@@ -142,24 +142,39 @@ public CountOptimisedDatabaseSubGraph(Transaction transaction, KernelTransaction
...
     @Override
     public long countsForRelationship(Label start, RelationshipType type) {
-        return read.countsForRelationship(
-                tokenRead.nodeLabel(start.name()), tokenRead.relationshipType(type.name()), TokenConstants.ANY_LABEL);
+        int relTypeID = tokenRead.relationshipType(type.name());
+        int nodeLabelID = tokenRead.nodeLabel(start.name());
+        if (nodeLabelID < 0 || relTypeID < 0) {
+            return 0;
+        }
+        return read.countsForRelationship(tokenRead.nodeLabel(start.name()), relTypeID, TokenConstants.ANY_LABEL);
```
hvub, line 169: Just for symmetry with the previous method:

```suggestion
        return read.countsForRelationship(nodeLabelID, relTypeID, TokenConstants.ANY_LABEL);
```

hvub (approved):
Excellent! 🎉

PR #746, opened 2025-02-28
Set up more robust virtual entities for deleted nodes and relationships

gem-neo4j:
When accessing a deleted relationship or node from a trigger, rel start and end node references still held onto the hold tx which caused errors and neither contained real ids and element ids which is unexpected behaviour

### DIFF
--- a/core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
+++ b/core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
@@ -30,6 +30,7 @@ import static org.neo4j.internal.helpers.collection.MapUtil.map;
 import static org.neo4j.test.assertion.Assert.assertEventually;
 
 import apoc.HelperProcedures;
+import apoc.convert.Json;
 import apoc.nodes.Nodes;
 import apoc.util.TestUtil;
 import apoc.util.Util;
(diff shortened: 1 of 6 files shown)


core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -143,6 +144,28 @@ public void testRemoveNode() {
...
+
+        String name = "access-removed-node";
+        String query = "CREATE (:TEST {data: apoc.convert.toJson($deletedNodes)})";
+        sysDb.executeTransactionally(
+                "CALL apoc.trigger.install('neo4j', $name, $query, {phase: 'afterAsync'})",
+                Map.of("name", name, "query", query));
+        awaitTriggerDiscovered(db, name, query);
+
+        db.executeTransactionally("MATCH (f:Foo) DETACH DELETE f");
+        Util.sleep(500);
```
loveleif, line 159: I would prefer to use the await until asserted thing, but we can always switch if it becomes flaky.

loveleif (approved):
Good find, looks solid!

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
gem-neo4j: Ah yes, I had made a mental note to do that, and forgot, so thanks! will fix up

PR #747, opened 2025-03-03
Deprecated hyperbolic functions moved into Cypher

gem-neo4j:
All of these are getting Cypher 25 support, so deprecate them :D


Lojjs (approved):
Seems good to me

PR #752, opened 2025-03-13
Bump dependencies

loveleif:


core/src/test/java/apoc/export/cypher/ExportCypherMultiRelTest.java
```diff
@@ -206,7 +207,7 @@ private void testsCommon(String expectedCypherStatement, Map<String, Object> oth
                 r -> Iterators.stream(r.<String>columnAs("cypherStatements")).collect(Collectors.joining("\n")));
 
         // check cypherStatements result
-        assertEquals(expectedCypherStatement, cypherStatements.replace("  ", " "));
+        assertThat(cypherStatements.replace("  ", " ")).isEqualTo(expectedCypherStatement);
```
loveleif, line 210: Ooops, not intentional, but don't hurt (actually better error messages).

core/src/test/java/apoc/agg/CollAggregationTest.java
loveleif, line 1: This file demonstrates how to use junit 5. All new tests should be made in this style and old tests can slowly be migrated.

gem-neo4j (approved):
LGTM :) Nice to start using enterprise! We can probably convert a bunch of our it tests now to simpler tests as previously docker was the only way for APOC to test enterprise features

PR #759, opened 2025-03-20
Move tests to block

JoelBergstrand:


core/src/test/java/apoc/export/csv/ExportCsvTest.java
```diff
@@ -211,16 +246,16 @@ public class ExportCsvTest {
...
+            3,Andrea,Milano,Via Garibaldi, 7,Address1;Address
             """;
     private static final String EXPECTED_ALL_NONE_ESCAPING =
             """
-            :ID,age:long,name,:LABEL
-            6,1,I am "groot", and more ,ESCAPING
-            7,2, ,ESCAPING
-            8,3,,ESCAPING
-            9,4,,ESCAPING
+            :ID,name,age:long,:LABEL
```
loveleif, line 254: I assume the ordering of these columns depends on some iteration order that changes in block?

loveleif (approved):
👍

core/src/test/java/apoc/export/csv/ExportCsvTest.java
JoelBergstrand: Yes sir!

Issue #159, opened 2022-09-01
Failed to invoke procedure `apoc.import.graphml`: Caused by: java.lang.RuntimeException: Error executing in separate thread


neo-technology-build-agent:
<a href="https://github.com/chrix75"><img src="https://avatars.githubusercontent.com/u/948863?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [chrix75](https://github.com/chrix75)**
_Tuesday Nov 08, 2016 at 19:51 GMT_
_Originally opened as https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/206_

----

Hi,

I've exported a graph with the procedure `apoc.export.graphml` but when I try to import it again I've got this error:

`Failed to invoke procedure `apoc.import.graphml`: Caused by: java.lang.RuntimeException: Error executing in separate thread
`

Thanks for your help.

neo-technology-build-agent:
<a href="https://github.com/jexp"><img src="https://avatars.githubusercontent.com/u/67427?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jexp](https://github.com/jexp)**
_Tuesday Apr 11, 2017 at 06:43 GMT_

----

@chrix75 can you try again with a more recent release?

neo-technology-build-agent:
<a href="https://github.com/MaryamMahmoodi"><img src="https://avatars.githubusercontent.com/u/1450796?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [MaryamMahmoodi](https://github.com/MaryamMahmoodi)**
_Saturday Jan 19, 2019 at 16:04 GMT_

----

@jexp 
Hi I have a problem like above
I have neo4j 3.4.10  and  apoc 3.4.0.4 , and I should import a graphml file that is created with apoc 3.5, but when import the file I got  this error on neo4j  :
Neo.ClientError.Procedure.ProcedureCallFailed: Failed to invoke procedure `apoc.import.graphml`: Caused by: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[3810,158]
Message: An invalid XML character (Unicode: 0x0) was found in the element content of the document.

 I need some help to resole that .

neo-technology-build-agent:
<a href="https://github.com/jexp"><img src="https://avatars.githubusercontent.com/u/67427?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jexp](https://github.com/jexp)**
[...]

PR #760, opened 2025-03-25
Fix issue with import cast exception

gem-neo4j:
Fixes: https://github.com/neo4j/apoc/issues/159


gem-neo4j:
APOC has no support or hint of support for default tags, it is hard to tell exactly what we should support or not, but I figured this fix can just ignore it as it assumed it wasn't there in the first place. From looking online, the usage of default is more to signify properties like colour, which are irrelevant to APOC's use case.

I am pretty sure there are more graphml bugs, as our implemented isn't very solid, but I'll leave that to another time :)

Issue #761, opened 2025-03-25
apoc.refactor.mergeNodes fails to combine empty list properties


michaelncy:
## Guidelines

Please note that GitHub issues are only meant for bug reports/feature requests. If you have questions on how to use APOC, please ask on [the Neo4j Discussion Forum](https://community.neo4j.com/c/neo4j-graph-platform/procedures-apoc) instead of creating an issue here.

## Expected Behavior (Mandatory)
`apoc.refactor.mergeNodes` should be able to combine empty list properties

## Actual Behavior (Mandatory)
When calling `apoc.refactor.mergeNodes` On nodes that have relationships that contain properties that are lists, with mergeRels: true and with the combine setting on these properties, if the relationships have a property as an empty list, it fails on: ```Failed to invoke procedure `apoc.refactor.mergeNodes`: Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0```

## How to Reproduce the Problem
Try to merge 2 nodes that have 2 relationships to another node where these relationships have a property that is an empty list. Pass mergeRels: true and 'combine' in the configuration to `apoc.refactor.mergeNodes`

### Simple Dataset (where it's possibile)

```
//Insert here a set of Cypher statements that helps us to reproduce the problem

MERGE (t:TEST)-[r:ACCESS]->(t2:BLA)<-[r2:ACCESS]-(t3:TEST) SET r.test = [], r2.test = []


MATCH (t:TEST)
WITH collect(t) AS tests
CALL apoc.refactor.mergeNodes(tests, {properties:"combine", mergeRels: true, singleElementAsArray: true}) YIELD node return node

```

### Steps (Mandatory)

  1. Create two nodes with relationships to another node that have a property that is an empty list
  2. Try to merge the two nodes

## Screenshots (where it's possibile)

<img width="845" alt="Image" src="https://github.com/user-attachments/assets/879408b8-9ffe-4d1c-847a-306ed9bd4a88" />

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: 
  - Neo4j:  "Neo4j Kernel" | "5.27-aura" | "enterprise"
  - Neo4j-Apoc: "2025.03.0"

gem-neo4j:
Hi there! Thanks for writing in :) I will fix this and it should be released in the next update (in a month or so I believe)

PR #762, opened 2025-03-26
Fix issue with empty arrays not being merged

gem-neo4j:
Fixes: https://github.com/neo4j/apoc/issues/761

### DIFF
--- a/core/src/main/java/apoc/refactor/util/PropertiesManager.java
+++ b/core/src/main/java/apoc/refactor/util/PropertiesManager.java
@@ -83,7 +83,9 @@ public class PropertiesManager {
     private static Object createPropertyValueFromSet(Set<Object> input, RefactorConfig refactorConfig) {
         Object array;
         try {
-            if (input.size() == 1 && !refactorConfig.isSingleElementAsArray()) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/refactor/util/PropertiesManager.java
```diff
@@ -83,7 +83,9 @@ public static void combineProperties(Map.Entry<String, Object> prop, Entity targ
     private static Object createPropertyValueFromSet(Set<Object> input, RefactorConfig refactorConfig) {
         Object array;
         try {
-            if (input.size() == 1 && !refactorConfig.isSingleElementAsArray()) {
+            if (input.isEmpty()) {
+                return new int[0];
```
loveleif, line 87: If I remember correctly we create empty list as empty string arrays in Cypher (might be wrong). I don't think it makes a difference, certainly not in driver at least, maybe in core API. We could align here, but I'm also fine with keeping this, just wanted to point this out.

gem-neo4j:
@loveleif were you happy with this otherwise? :)

JoelBergstrand (approved):

PR #766, opened 2025-03-31
Allow null values for apoc.util.validatePredicate

gem-neo4j:
Allow null values as implied by the documentation.


loveleif (approved):
Great!

Issue #769, opened 2025-04-07
Missing relationships when restoring database from apoc.export.cypher.all dumps


vetlesim:
## Expected Behavior
Restoring a database from a dump generated by `apoc.export.cypher.all` should preserve relationships between nodes, even when those nodes have undefined properties involved in a composite property uniqueness constraint.

## Actual Behavior
By default, `apoc.export.cypher.all` relies on existing uniqueness constraints to match nodes when constructing relationships. However, if any of the properties in the composite constraint are undefined on a node, the match silently fails and the relationship is not created when later restoring from the dump.

### Simple Dataset

```
CREATE CONSTRAINT composite_unique
FOR (n:Node)
REQUIRE (n.propA, n.propB) IS UNIQUE;

CREATE (:Node {propA: "X", propB: "1"});
CREATE (:Node {propA: "Y"});

MATCH (a:Node {propA: "X", propB: "1"})
MATCH (b:Node {propA: "Y"})
CREATE (a)-[:LINKED_TO]->(b);
```

### Steps
1. Create two nodes and a relationship between them using the cypher statements above. The node properties `propA` and `propB` are subject to a composite property uniqueness constraint.
2. Make a dump of the database
`CALL apoc.export.cypher.all('dump.cypher', {writeNodeProperties:true})`
3. Wipe the database
`MATCH (n) DETACH DELETE n`
5. Re-import from the dump
`CALL apoc.cypher.runFile('dump.cypher')`

Observe how the two nodes no longer are connected with the `LINKED_TO` relationship.

## Specifications

Currently used versions

### Versions

  - OS: Debian 11 (Bullseye) with Docker
  - Neo4j: 5.19.0
  - Neo4j-Apoc: 5.19.0
  - Neo4j-Apoc-Extended: 5.19.0

JoelBergstrand:
Hi @vetlesim, thank you for reporting this issue and the well written bug report. I've confirmed the issue on our latest release as well, the team will look into this issue as soon as possible.

PR #772, opened 2025-04-22
Fix issue with export with constraints

gem-neo4j:
Fixes https://github.com/neo4j/apoc/issues/769 by making sure that if a unique constraint is used, then the node has all the required unique properties


gem-neo4j:
https://trello.com/c/Dp8QmE2S/2645-missing-relationships-when-restoring-database-from-apocexportcypherall-dumps

loveleif (approved):
Great!

I had a bit of a hard time understanding what's going on here. Could there be another bug with multiple indexes, looks like we pick the first constraint, but multiple could match, right? In situations like: 

```
create constraint for (n:A) require (n.a,n.b) is unique;
create constraint for (n:A) require (n.c, n.d, n.e) is unique;
create constraint for (n:B) require (n.a, n.c) is unique;

create (:A:B {a:1, b:2, c:3, d:4, e:5})
```

gem-neo4j:
@loveleif yeah, but it shouldn't matter which one we pick as matching on any will provide the correct node if a and b are unique on label A then filtering on a and b (assuming a and b aren't null, which this fix checks) will find the one node that matches that, and that node is the same one as if we had searched for it with one of the other constraints

PR #776, opened 2025-05-05
Make periodic procedures safe to call concurrently

loveleif:
https://trello.com/c/Vi3h8rVw/2396-we-should-not-allow-concurrent-execution-of-apocperiodicrepeat


gem-neo4j (approved):
LGTM :)

PR #777, opened 2025-05-05
Migrate test to Junit 5

loveleif:


core/src/test/java/apoc/export/csv/ExportCsvTest.java
```diff
@@ -423,31 +474,48 @@ public class ExportCsvTest {
...
-            """;
-    private static final String EXPECTED_ALL_DIFFERENTIATE_NULLS_IF_NEEDED_REL =
+            """,
+            """
+            :ID,age:long,name,:LABEL
+            6,1,"I am ""groot"", and more ",ESCAPING
+            7,2, ,ESCAPING
+            8,3,"",ESCAPING
+            9,4,,ESCAPING
+            """);
```
JoelBergstrand, line 509: Are these possible different orderings of the result?

JoelBergstrand: Are we sure there can't be more different orderings?

JoelBergstrand (approved):

JoelBergstrand:
What is the failure?

core/src/test/java/apoc/export/csv/ExportCsvTest.java
loveleif: No, if this breaks it will be better to parse the csv.

loveleif:
> What is the build failure?

Unrelated I believe

PR #787, opened 2025-05-16
Bump dependencies

loveleif:
Fixes some flaky tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,10 @@ subprojects {
     dependencies {
         testImplementation(platform("org.junit:junit-bom:5.11.4"))
         testImplementation('org.junit.jupiter:junit-jupiter')
+
+        // We want the same netty version as other neo4j dependencies, are there better ways than to keep this in sync?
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -59,6 +59,10 @@ subprojects {
     dependencies {
         testImplementation(platform("org.junit:junit-bom:5.11.4"))
         testImplementation('org.junit.jupiter:junit-jupiter')
+
+        // We want the same netty version as other neo4j dependencies, are there better ways than to keep this in sync?
+        implementation(enforcedPlatform('io.netty:netty-bom:4.2.1.Final'))
```
loveleif, line 64: No idea what I'm doing. Are there better ways?

JoelBergstrand: Will this start failing if there is a different version in the monorepo?

loveleif: Not as far as I know, it will force the netty version to be consistent though I believe (netty consists of multiple dependencies).

PR #789, opened 2025-05-19
Fix flaky tests

loveleif:


core/src/test/java/apoc/periodic/PeriodicTest.java
```diff
@@ -195,7 +192,15 @@ public void testSubmitStatement() throws Exception {
...
-        testCall(db, callList, (r) -> assertEquals(true, r.get("done")));
+        try (final var tx = db.beginTx();
+                final var result = tx.execute(callList)) {
+            final var resultList = result.stream().toList();
+            // We clean up completed tasks periodically so this can be empty
+            if (!resultList.isEmpty()) {
+                assertThat(resultList)
+                        .contains(Map.of("name", "foo", "done", true, "cancelled", false, "delay", 0L, "rate", 0L));
+            }
+        }
```
loveleif, line 203: Question: Why was this not flaky before?
Answer: The new test framework do not restart the dbms (only the db), so we might cleanup the task (it's done every 10s).

gem-neo4j (approved):
Great :) thanks for the question and answer already supplied!

PR #791, opened 2025-06-04
Deprecate conditional based procedures in Cypher 25 and recommend Conditional queries instead

gem-neo4j:
When using Cypher 25, APOC should give a warning about deprecation on the conditional based procedures, the docs will also be updated to show the deprecation as well as give examples on how to actually use the new conditional queries


loveleif (approved):
👍

PR #793, opened 2025-06-11
Deprecate procedures and functions that are easy to do in Cypher and are more performant in Cypher as well.

gem-neo4j:
Based on the benchmarking analysis, these are the procedures and functions that I deemed were easy to do in Cypher as well as showed a better performance in Cypher.

See [here](https://github.com/neo4j/docs-apoc/pull/439) for the docs

### DIFF
--- a/core/src/main/java/apoc/nodes/Nodes.java
+++ b/core/src/main/java/apoc/nodes/Nodes.java
@@ -223,6 +223,7 @@ public class Nodes {
         return delete(nodes, batchSize);
     }
 
+    @Deprecated
     @Procedure(name = "apoc.nodes.delete", mode = Mode.WRITE, deprecatedBy = "Cypher's `CALL {...} IN TRANSACTIONS`.")
     @QueryLanguageScope(scope = {QueryLanguage.CYPHER_25})
     @Description("Deletes all `NODE` values with the given ids.")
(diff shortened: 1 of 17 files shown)


core/src/main/java/apoc/nodes/Nodes.java
```diff
@@ -208,16 +210,31 @@ public Stream<NodeResult> get(
...
+                            value = "nodes",
+                            description =
+                                    "The nodes to be deleted. Nodes can be of type `STRING` (elementId()), `INTEGER` (id()), `NODE`, or `LIST<STRING | INTEGER | NODE>`.")
+                    Object nodes,
+            @Name(value = "batchSize", description = "The number of node values to delete in a single batch.")
+                    long batchSize) {
+        return delete(nodes, batchSize);
+    }
+
+    @Procedure(name = "apoc.nodes.delete", mode = Mode.WRITE, deprecatedBy = "Cypher's `CALL {...} IN TRANSACTIONS`.")
```
loveleif, line 226: Missing a `@Deprecated`?

loveleif: The dynamic batch size can't be used in cypher, but feels like we can live with that.

loveleif (approved):
Excellent work :)

core/src/main/java/apoc/nodes/Nodes.java
gem-neo4j: Yeah, we seem to have been okay with that from the initial CIT analysis, I just extrapolated from there 👀

Issue #444, opened 2023-06-24
Failed import graphml due to 'nan' input string


MikeB2019x:
I have been using the following command to import `.graphml` files in to neo4j
```
CALL apoc.import.graphml("test.graphml", {readLabels: true, storeNodeIds:true})
```
Recently I began working with larger files (3GB) derived from larger data tables and started getting this error when trying to import.  
```
Failed to invoke procedure `apoc.import.graphml`: Caused by: java.lang.NumberFormatException: For input string: "nan"
```  
Yes there are `<data key=xxx > nan</data>` entries in the graphml b/c it was derived from a table with NaN values.  Is there a workaround?  Note that the import will work if I replace all the `nan` with no character i.e. `<data></data>` but that seems an unecessary hack.

I'm am working on creating a small data set but my workflow is:

  1.  Import tabular data from postgres db
  1. Create a graph using `pandas` and `networkx` (incl. creating a proper 'labels' as a node attribute)
  1. From networkx export a `.graphml`
  1. Import `.graphml` in to neo4j with command indicated at beginning 

## Screenshots
![image](https://github.com/neo4j/apoc/assets/49003263/efc3f21a-6eac-4425-9561-789464d8c71b)

### Versions

  - OS:  Mac Pro M1 w/ Ventrua 13.3.1 (a)
  - Neo4j: 4.4.o (community)
  - Neo4j-Apoc: 4.4.0.1

gem-neo4j:
Hi, thank you writing in :) I will ticket this, however, we currently have APOC under maintenance mode so cannot say when this would be possible for us to put in, feel free to make a PR yourself for us to review if you want it done faster! Thanks

gem-neo4j:
So we are focusing on only fixing bugs and security issues and not adding new functionality atm :)

hvub:
Cardified. Closing here.

PR #798, opened 2025-07-03
Add NAN support to graphml imports

gem-neo4j:
Fixes: https://github.com/neo4j/apoc/issues/444

Gives support for import Graphml with NAN values for number types. Int and Long in Java do not support NaN, so it will be a Float.NaN in this case, the only potential issue is if a user has a constraint on what they are importing for having it as an INTEGER, but I believe this is a better issue than the import just failing with an exception.


hvub (approved):
🎉

The documentation should point out explicitly that NANs are always imported as Floats regardless of the `attr.type` in `<key ...>` element of the imported GraphML file.

gem-neo4j:
Docs: https://github.com/neo4j/docs-apoc/pull/448

PR #805, opened 2025-07-16
Migrate from deprecated neo4j-java-driver methods to new ones

injectives:
The main objective is to stop using deprecated methods that have been removed in the next major driver version.


Lojjs (approved):
Thanks for fixing!

PR #827, opened 2025-08-19
Fix flaky tests

loveleif:


core/src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -392,12 +398,12 @@ public void testRelationshipExists() {
         db.executeTransactionally("CREATE INDEX rel_index_composite FOR ()-[r:PURCHASED]-() ON (r.date, r.amount)");
         awaitIndexesOnline();
 
-        assertTrue(singleResultFirstColumn(db, "RETURN apoc.schema.relationship.indexExists('KNOWS', ['since'])"));
```
hvub, line 395: Why is this assertion removed? It feels like is should still be true, doesn't it?

core/src/test/java/apoc/schema/SchemasTest.java
```diff
@@ -843,9 +849,9 @@ public void testIndexesMoreLabels() {
                         assertEquals("NO FAILURE", r.get("failure"));
                         assertEquals(100d, r.get("populationProgress"));
                         assertEquals(1d, r.get("valuesSelectivity"));
-                        Assertions.assertThat(r.get("userDescription").toString())
+                        assertThat(r.get("userDescription").toString())
                                 .contains(
-                                        "name='index3', type='TEXT', schema=(:Person {name}), indexProvider='text-2.0' )");
+                                        "name='index3', type='TEXT', schema=(:Person {name}), indexProvider='text-3.0' )");
```
hvub, line 854: Not sure it is make sense to assert on a specific index provider.

hvub (approved):
Looks good.

core/src/test/java/apoc/schema/SchemasTest.java
loveleif: Agree, but I'll leave that for the next time :).

core/src/test/java/apoc/schema/SchemasTest.java
loveleif: Fixed: 3cb7e8f

PR #832, opened 2025-08-27
Update startup test to not include arrow procedures in community

Lojjs:
The dependencies for arrow are not bundled with APOC but needs to be provided explicitly. Before they were provided through both Neo4j enterprise and community, but now they are only in enterprise on 5.26. This means the arrow procedures are no longer packaged by default in a community docker image.


Lojjs:
This was just for testing

PR #845, opened 2025-09-25
Update method call from fulltext to forSemanticSearch

parnmatt:
This particular schema descriptor will soon no longer be specific to the
fulltext index.

We have renamed (and aliased) this method and will be removing the
internal fulltext method alias after this is approved and merged.


loveleif (approved):
Thanks!

PR #848, opened 2025-09-25
Update generated constraint name

parnmatt:
There has been a change in the schema rule generated name algorithm.


loveleif (approved):
Thanks, not sure why this test have assertions on the generated name, but looks innocent enough.

parnmatt:
> Thanks, not sure why this test have assertions on the generated name, but looks innocent enough.

Yes, I would prefer as few tests as possible to have those assertions. Feel free to remove that section of the test in the future; the relationship version doesn't have it (so there's precedence).

PR #852, opened 2025-10-01
Update schema descriptor for disjunction of entity tokens for fulltext

parnmatt:
`FULLTEXT` index descriptor strings now reflect the disjunction of multiple entry tokens.
Rather than: `(:Label1:Label2:Label3 {prop1, prop2, prop3})`  
It should be: `(:Label1|Label2|Label3 {prop1, prop2, prop3})`


loveleif (approved):
Thanks!

PR #854, opened 2025-10-06
Updated manifest with more information

JoelBergstrand:
Updated APOC manifests to include vendor and title.


gem-neo4j (approved):
This looks fine to me, extended uses common and shouldn't use the core gradle file, and I checked, extended has its own manifest, so I don't see this being an issue

PR #856, opened 2025-10-13
Vector indexes

gem-neo4j:
Remove vector flag as well as fix up vector index support


core/src/main/java/apoc/schema/Schemas.java
```diff
@@ -200,6 +200,8 @@ public List<AssertSchemaResult> assertIndexes(
 
         for (IndexDefinition definition : Util.getIndexes(tx)) {
             if (definition.getIndexType() == IndexType.LOOKUP) continue;
+            // Don't drop vector indexes
+            if (definition.getIndexType() == IndexType.VECTOR) continue;
```
loveleif: Why not?

loveleif (approved):
Great! Sorry for the delay.

PR #860, opened 2025-11-05
Migrate junit version in tests

gem-neo4j:
Migrating tests away from 4, I followed the 5 migration, but from what I can see it isn't as big of a switch to 6, so should be easier to migrate there :)


loveleif (approved):
Nice progress!

PR #864, opened 2025-11-19
Fix up isDense function for block format

gem-neo4j:
Fix isDense for block format, this adds an optional argument which allows for accurate results on block format (where previously true would always be returned).

Docs: https://github.com/neo4j/docs-apoc/pull/482

Linear: resolves SURF-261


loveleif (approved):
Nice

PR #865, opened 2025-11-19
Fix periodic submit flaky test

gem-neo4j:
Apply a wait and try method instead of just waiting 1s like other tests in the file do

Linear: resolves SURF-244, SURF-164


loveleif (approved):
Great!

PR #884, opened 2026-01-19
Migrate more junit tests

gem-neo4j:
Migrate the rest of the Junit4 tests over to Junit5. This PR doesn't remove Junit4 yet from the dependencies, I want to get this one merged and then see if I can remove it in a follow up (also will need to make checks on APOC extended when doing so)

### DIFF
--- a/common/src/test/java/apoc/util/kernel/MultiThreadedGlobalGraphOperationsTest.java
+++ b/common/src/test/java/apoc/util/kernel/MultiThreadedGlobalGraphOperationsTest.java
@@ -31,7 +31,7 @@ import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.test.extension.Inject;
 
 @ImpermanentEnterpriseDbmsExtension(createDatabasePerTest = false)
-public class MultiThreadedGlobalGraphOperationsTest {
+class MultiThreadedGlobalGraphOperationsTest {
 
     @Inject
     private GraphDatabaseAPI db;
--- a/common/src/test/java/apoc/util/kernel/MultiThreadedGlobalGraphOperationsTest.java
+++ b/common/src/test/java/apoc/util/kernel/MultiThreadedGlobalGraphOperationsTest.java
@@ -20,39 +20,30 @@ package apoc.util.kernel;
 
 import static apoc.util.kernel.MultiThreadedGlobalGraphOperations.BatchJobResult;
 import static apoc.util.kernel.MultiThreadedGlobalGraphOperations.forAllNodes;
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import com.neo4j.test.extension.ImpermanentEnterpriseDbmsExtension;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
-import org.junit.ClassRule;
-import org.junit.Test;
-import org.neo4j.test.rule.DbmsRule;
-import org.neo4j.test.rule.ImpermanentDbmsRule;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.test.extension.Inject;
 
+@ImpermanentEnterpriseDbmsExtension(createDatabasePerTest = false)
 public class MultiThreadedGlobalGraphOperationsTest {
 
-    @ClassRule
-    public static DbmsRule db = new ImpermanentDbmsRule();
+    @Inject
+    private GraphDatabaseAPI db;
 
-    @BeforeClass
-    public static void beforeClass() {
-        createData();
-    }
-
-    @AfterClass
-    public static void teardown() {
-        db.shutdown();
-    }
-
-    private static void createData() {
+    @BeforeAll
+    void beforeClass() {
         db.executeTransactionally(
                 "UNWIND range(1,1000) as x MERGE (s{id:x}) MERGE (e{id:x+1}) merge (s)-[:REL{id:x}]->(e)");
     }
 
     @Test
-    public void shouldforAllNodesWork() {
+    void shouldWorkForAllNodes() {
         AtomicInteger counter = new AtomicInteger();
         BatchJobResult result =
                 forAllNodes(db, Executors.newFixedThreadPool(4), 10, (nodeCursor) -> counter.incrementAndGet());
--- a/core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
+++ b/core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
@@ -36,7 +36,7 @@ import org.neo4j.test.extension.ExtensionCallback;
 import org.neo4j.test.extension.Inject;
 
 @ImpermanentEnterpriseDbmsExtension(configurationCallback = "configure")
-public class PathsToJsonTreeTest {
+class PathsToJsonTreeTest {
 
     @Inject
     GraphDatabaseService db;
@@ -52,7 +52,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeSimplePath() {
+    void testToTreeSimplePath() {
         /*            r:R
               a:A --------> b:B
         */
@@ -93,7 +93,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testSingleNode() {
+    void testSingleNode() {
         // a:A
         db.executeTransactionally("CREATE (a: A {nodeName: 'a'})");
 
@@ -120,7 +120,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testSingleDisjointNodes() {
+    void testSingleDisjointNodes() {
         // a:A
         db.executeTransactionally("CREATE (a: A {nodeName: 'a'}), (b: B {nodeName: 'b'}), (c: C {nodeName: 'c'})");
 
@@ -168,7 +168,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeSimpleReversePath() {
+    void testToTreeSimpleReversePath() {
         /*            r:R
               a:A <-------- b:B
         */
@@ -208,7 +208,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeSimpleBidirectionalPath() {
+    void testToTreeSimpleBidirectionalPath() {
         /*         r1:R
                  -------->
              a:A          b:B
@@ -267,7 +267,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeSimpleBidirectionalQuery() {
+    void testToTreeSimpleBidirectionalQuery() {
         /*         r1:R
              a:A --------> b:B
         */
@@ -309,7 +309,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeSimpleQueryDisjointPaths() {
+    void testToTreeSimpleQueryDisjointPaths() {
         /*         r1:R
              a:A --------> b:B
         */
@@ -371,7 +371,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeBidirectionalPathAndQuery() {
+    void testToTreeBidirectionalPathAndQuery() {
         /*          r1:R1         r2:R2
               a:A ---------> b:B --------> a
         */
@@ -426,7 +426,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeComplexGraph() {
+    void testToTreeComplexGraph() {
         /*          r1:R1         r2:R2
               a:A --------> b:B ------> c:C
                              |
@@ -494,7 +494,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeComplexGraphBidirectionalQuery() {
+    void testToTreeComplexGraphBidirectionalQuery() {
         /*          r1:R1         r2:R2
               a:A --------> b:B -------> c:C
                              |
@@ -563,7 +563,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeGraphWithLoops() {
+    void testToTreeGraphWithLoops() {
         /*          r1:R1          r2:R2
               a:A ---------> b:B --------> c:C
                             /  /|
@@ -630,7 +630,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testIncomingRelationships() {
+    void testIncomingRelationships() {
         /*          r1:R1         r2:R2
               a:A --------> b:B <------ c:C
         */
@@ -692,7 +692,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeMultiLabelFilters() {
+    void testToTreeMultiLabelFilters() {
         /*            r:R
               a:A:B -------> c:C
         */
@@ -733,7 +733,7 @@ public class PathsToJsonTreeTest {
     }
 
     @Test
-    public void testToTreeMultiLabelFiltersForOldProcedure() {
+    void testToTreeMultiLabelFiltersForOldProcedure() {
         /*            r:R
               a:A:B -------> c:C
         */
--- a/core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
+++ b/core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
@@ -35,7 +35,7 @@ import org.neo4j.test.TestDatabaseManagementServiceBuilder;
 import org.neo4j.test.extension.ExtensionCallback;
 import org.neo4j.test.extension.Inject;
 
-@ImpermanentEnterpriseDbmsExtension(createDatabasePerTest = true, configurationCallback = "configure")
+@ImpermanentEnterpriseDbmsExtension(configurationCallback = "configure")
 public class PathsToJsonTreeTest {
 
     @Inject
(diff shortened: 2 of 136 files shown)


common/src/test/java/apoc/util/kernel/MultiThreadedGlobalGraphOperationsTest.java
```diff
@@ -20,39 +20,30 @@
...
-import org.junit.Test;
-import org.neo4j.test.rule.DbmsRule;
-import org.neo4j.test.rule.ImpermanentDbmsRule;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.test.extension.Inject;
 
+@ImpermanentEnterpriseDbmsExtension(createDatabasePerTest = false)
 public class MultiThreadedGlobalGraphOperationsTest {
```
Lojjs: Optional cleanup:
You could remove “public” from the test classes, @BeforeAll methods and individual tests more consistently. I think it is not necessary to get it to work but my Intellij shows them as warnings.

core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
```diff
@@ -35,7 +35,7 @@
 import org.neo4j.test.extension.ExtensionCallback;
 import org.neo4j.test.extension.Inject;
 
-@ImpermanentEnterpriseDbmsExtension(createDatabasePerTest = true, configurationCallback = "configure")
```
Lojjs, line 38: Are we sure it is safe to remove `createDatabasePerTest = true` here as there are multiple tests creating data in the db? Seems to go green locally, but I'm not sure if TC runs it more in parallel

it/src/test/java/apoc/it/core/ExportJsonS3Test.java
```diff
@@ -30,39 +30,48 @@
...
+    @Inject
+    GraphDatabaseService db;
 
-    @Before
-    public void setUp() {
+    @BeforeAll
+    void setUpAll() {
+        // Initialize S3 test container and APOC config
+        S3BaseTest.baseBeforeClass();
         apocConfig().setProperty(APOC_EXPORT_FILE_ENABLED, true);
```
Lojjs: Do we need this line in setupAll() as we are anyway doing it for each test in setup()?

core/src/test/java/apoc/trigger/TriggerTest.java
```diff
@@ -18,75 +18,93 @@
...
+        System.setProperty("apoc.trigger.enabled", "true");
+        builder.setConfigRaw(Map.of("internal.dbms.debug.track_cursor_close", "true"));
+        builder.setConfig(GraphDatabaseSettings.default_language, GraphDatabaseSettings.CypherVersion.Cypher5);
+    }
 
-    @Before
-    public void setUp() {
+    @BeforeAll
+    void beforeAll() {
         start = System.currentTimeMillis();
```
Lojjs, line 76: Before, `start` was the starting time of each test run, now it is the starting time of the whole test file run. This might impact the assertions where we check `getProperty("txTime") > start` somewhat, as now a time before the test was started would also be considered OK. I don't really know these tests well enough to judge if that is a problem though.

core/src/test/java/apoc/trigger/TriggerTest.java
```diff
@@ -18,75 +18,93 @@
...
+import org.neo4j.test.TestDatabaseManagementServiceBuilder;
+import org.neo4j.test.extension.ExtensionCallback;
+import org.neo4j.test.extension.Inject;
 
 /**
  * CYPHER 5 only; moved to extended for Cypher 25
  */
+@EnterpriseDbmsExtension(configurationCallback = "configure", createDatabasePerTest = false)
 public class TriggerTest {
+    private static final File directory = new File("target/conf");
```
Lojjs: How come the directory thing is needed now? It seems to work without it locally for me but I think I am on junit5 and not junit6

core/src/test/java/apoc/trigger/TriggerTest.java
```diff
@@ -99,7 +117,7 @@ public void testListTriggers() {
     }
 
     @Test
-    public void testRemoveNode() {
+    void testRemoveNode() {
```
Lojjs, line 118: I tried to run the suite locally to understand some of the setup. On the first run this test failed, but I have run it again like 10 times and cannot reproduce it, so either my Intellij is just confusing me or it is flaky

core/src/test/java/apoc/spatial/SpatialTest.java
```diff
@@ -322,27 +332,26 @@ public void testAllTheThings() {
         testResult(db, query, res -> {
             int left = expectedCount;
             while (left > 0) {
-                assertTrue(
-                        "Expected " + expectedCount + " results, but got only " + (expectedCount - left),
-                        res.hasNext());
+                assertTrue(res.hasNext());
```
Lojjs, line 335: Not sure if there is anything we can do about it, but it is a bit sad that we no longer get the extra info about what `left` was in case this test would go flaky in the future.

core/src/test/java/apoc/export/json/ExportJsonTest.java
```diff
@@ -77,19 +80,23 @@ public class ExportJsonTest {
...
+        // Run with aligned format to get sequential ids (assertions depends on this)
+        builder.setConfig(GraphDatabaseSettings.db_format, "aligned");
+    }
+
+    @BeforeEach
     public void setup() {
-        TestUtil.registerProcedure(db, ExportJson.class, ImportJson.class, Graphs.class, HelperProcedures.class);
         apocConfig().setProperty(APOC_IMPORT_FILE_ENABLED, true);
         apocConfig().setProperty(APOC_EXPORT_FILE_ENABLED, true);
+        db.executeTransactionally("MATCH (n) DETACH DELETE n");
```
Lojjs: This line should not be needed as we create a new db for each test.

core/src/test/java/apoc/export/graphml/ExportGraphMLTestUtil.java
```diff
@@ -253,4 +253,23 @@ public static void setUpGraphMl(GraphDatabaseService db, TestName testName) {
         db.executeTransactionally(
                 "CREATE (f:Foo:Foo2:Foo0 {name:'foo', born:Date('2018-10-10'), place:point({ longitude: 56.7, latitude: 12.78, height: 100 })})-[:KNOWS]->(b:Bar {name:'bar',age:42, place:point({ longitude: 56.7, latitude: 12.78})}),(c:Bar {age:12,values:[1,2,3]})");
     }
+
+    public static void setUpGraphMl(GraphDatabaseService db, String methodName) {
+        setUpGraphMlDataOnly(db, methodName);
+    }
+
+    public static void setUpGraphMlDataOnly(GraphDatabaseService db, String methodName) {
```
Lojjs: Can we not just merge these methods as the first one just calls the second with the exact same arguments? Also the method above now seems unused.

core/src/test/java/apoc/export/csv/ExportCsvNeo4jAdminTest.java
```diff
@@ -100,16 +103,24 @@ public class ExportCsvNeo4jAdminTest {
...
-    public void before() {
+    @ExtensionCallback
+    void configure(TestDatabaseManagementServiceBuilder builder) {
+        builder.setConfig(
+                GraphDatabaseSettings.load_csv_file_url_root, directory.toPath().toAbsolutePath());
+    }
+
+    @BeforeAll
+    void setUpAll() {
         apocConfig().setProperty(APOC_EXPORT_FILE_ENABLED, true);
```
Lojjs:
```suggestion
```

Again, I think there is no need to do this in both setUpAll() and setUp()

core/src/test/java/apoc/export/cypher/ExportCypherTest.java
```diff
@@ -97,35 +98,38 @@ public class ExportCypherTest {
...
+    GraphDatabaseService db;
 
-    @Rule
-    public TestName testName = new TestName();
+    @ExtensionCallback
+    void configure(TestDatabaseManagementServiceBuilder builder) throws java.io.IOException {
+        builder.setConfig(
+                GraphDatabaseSettings.load_csv_file_url_root,
+                directory.getCanonicalFile().toPath());
+        builder.setConfig(GraphDatabaseSettings.db_format, "aligned");
```
Lojjs, line 110:
```suggestion
        // Test assertions depends on sequential ids
        builder.setConfig(GraphDatabaseSettings.db_format, "aligned");
```

not crucial but could be good to keep the comment about why we are testing on aligned

core/src/test/java/apoc/export/graphml/ExportGraphMLTest.java
```diff
@@ -90,30 +92,34 @@ public class ExportGraphMLTest {
...
     }
 
-    @Before
-    public void setUp() {
-        setUpGraphMl(db, testName);
+    @BeforeAll
+    void setUpAll() {
+        TestUtil.registerProcedure(db, ExportGraphML.class, Graphs.class, HelperProcedures.class);
+        apocConfig().setProperty(APOC_IMPORT_FILE_ENABLED, true);
+        apocConfig().setProperty(APOC_EXPORT_FILE_ENABLED, true);
```
Lojjs:
```suggestion
```

I don't think we need these lines here as we are overriding them in setupGraphMlDataOnly() for each test

core/src/test/java/apoc/export/json/ExportJsonTest.java
Lojjs:
```suggestion
```

This line should not be needed as we create a new db for each test.

core/src/test/java/apoc/export/json/ExportJsonTest.java
```diff
@@ -109,13 +116,15 @@ public void setup() {
...
     }
 
-    @After
-    public void teardown() {
-        db.shutdown();
+    @BeforeAll
+    void setUpAll() {
+        TestUtil.registerProcedure(db, ExportJson.class, ImportJson.class, Graphs.class, HelperProcedures.class);
+        apocConfig().setProperty(APOC_IMPORT_FILE_ENABLED, true);
+        apocConfig().setProperty(APOC_EXPORT_FILE_ENABLED, true);
```
Lojjs, line 120:
```suggestion
```

I don't think we need these here as we already have them in setUp()

core/src/test/java/apoc/path/ExpandPathTest.java
```diff
@@ -65,19 +70,14 @@ public static void setUp() {
...
-        db.shutdown();
-    }
-
-    @After
-    public void removeOtherLabels() {
+    @AfterEach
+    void removeOtherLabels() {
         db.executeTransactionally(
                 "OPTIONAL MATCH (c:Western) REMOVE c:Western WITH DISTINCT 1 as ignore OPTIONAL MATCH (c:Denylist) REMOVE c:Denylist");
     }
```
Lojjs:
```suggestion
```

Since we are now running each test in a new db, I think this is not needed anymore.

it/src/test/java/apoc/it/core/ExportCsvS3Test.java
```diff
@@ -22,41 +22,46 @@
...
+    @Inject
+    GraphDatabaseService db;
 
-    @BeforeClass
-    public static void setUp() {
+    @BeforeAll
+    void setUpAll() {
         baseBeforeClass();
-
         apocConfig().setProperty(APOC_EXPORT_FILE_ENABLED, true);
```
Lojjs:
```suggestion
```

Same as in some other tests, we do not need this line in both setUpAll() and setUp()

it/src/test/java/apoc/it/core/ExportJsonS3Test.java
Lojjs:
```suggestion
```

Same as in some other tests, we do not need this line in both setUpAll() and setUp()

core/src/test/java/apoc/warmup/WarmupTest.java
```diff
@@ -18,50 +18,43 @@
...
-                + "MATCH (n)-[r:KNOWS]->(m) "
-                + "WHERE NOT id(r) IN ids "
-                + "DELETE n, m, r");
+    @Inject
+    private GraphDatabaseService db;
+
+    @ExtensionCallback
+    void configure(TestDatabaseManagementServiceBuilder builder) {
+        builder.setConfig(GraphDatabaseSettings.default_language, GraphDatabaseSettings.CypherVersion.Cypher5)
+                .setConfig(GraphDatabaseSettings.db_format, "aligned");
```
Lojjs, line 53:
```suggestion
        builder.setConfig(GraphDatabaseSettings.default_language, GraphDatabaseSettings.CypherVersion.Cypher5)
                //  Procedure is only supported on record storage databases
                .setConfig(GraphDatabaseSettings.db_format, "aligned");
```

I think this comment is good to keep for context

core/src/test/java/apoc/warmup/WarmupTest.java
```diff
@@ -18,50 +18,43 @@
...
 
-    @Before
-    public void setUp() {
-        TestUtil.registerProcedure(db, Warmup.class);
-        // Create enough nodes and relationships to span 2 pages
-        db.executeTransactionally("CREATE CONSTRAINT FOR (f:Foo) REQUIRE f.foo IS UNIQUE");
-        db.executeTransactionally(
-                "UNWIND range(1, 300) AS i CREATE (n:Foo {foo:i})-[:KNOWS {bar:2}]->(m {foobar:3, array:range(1,100)})");
-        // Delete all relationships and their nodes, but ones with the minimum and maximum relationship ids, so
-        // they still span 2 pages
```
Lojjs, line 55: Do we want to transfer the comments from here to prepareData() or did you make a conscious decision to remove them? I guess they might be a bit helpful if someone wants to understand why the setup looks like it does.

core/src/test/java/apoc/trigger/TriggerTest.java
```diff
@@ -18,75 +18,93 @@
...
+    @Inject
+    DatabaseManagementService dbms;
+
+    static { //noinspection ResultOfMethodCallIgnored
+        directory.mkdirs();
+    }
+
+    @ExtensionCallback
+    void configure(TestDatabaseManagementServiceBuilder builder) {
+        System.setProperty("apoc.trigger.refresh", String.valueOf(TRIGGER_DEFAULT_REFRESH));
```
Lojjs: Out of curiosity, what is the purpose of adding this line, which we seem not to have had before?

core/src/test/java/apoc/export/ExportCoreSecurityTest.java
```diff
@@ -119,282 +117,208 @@ public static List<Object[]> getParameterData(
...
-            this.apocProcedure = getApocProcedure(exportMethod, exportMethodType, exportMethodArguments);
-            this.fileName = fileName;
-            this.consumer = consumer;
-        }
-
-        /*
-         * These test cases attempt to access a directory with the same prefix as the import directory. This is design to
-         * test "directoryName.startsWith" logic which is a common path traversal bug.
-         * All these tests should fail because they access a directory which isn't the configured directory
-         */
```
Lojjs, line 155: Do you think this comment would be worth to keep, or is it not very helpful?

core/src/test/java/apoc/export/ExportCoreSecurityTest.java
```diff
@@ -119,282 +117,208 @@ public static List<Object[]> getParameterData(
...
-
-        public static final List<String> casesOutsideDir =
-                Arrays.asList(case01, case02, case03, case04, case05, case07, case08);
-
-        /*
-        All of these will resolve to a local path after normalization which will point to
-        a non-existing directory in our import folder: /apoc. Causing them to error that is
-        not found. They all attempt to exit the import folder back to the apoc folder:
-        Directory Layout: .../apoc/core/target/import
-        */
```
Lojjs, line 176: Do you think this comment would be worth to keep, or is it not very helpful?

Lojjs (commented):
So nice that you finished this task 🎉 Only some minor cleanup suggestions and questions

core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
gem-neo4j: createDatabasePerTest = true is the default, so it only needs to be specified if it is false (My intellij was telling me to remove the unused code :P )

core/src/test/java/apoc/export/graphml/ExportGraphMLTestUtil.java
gem-neo4j: Nice spotting :)

core/src/test/java/apoc/convert/PathsToJsonTreeTest.java
Lojjs: Right, I missed that. All fine then

core/src/test/java/apoc/trigger/TriggerTest.java
gem-neo4j: Oh this test took so long to work, I ended up copying over the entire working one and just removing stuff till it worked 😅 but obviously forgot this line! not needed :)

core/src/test/java/apoc/trigger/TriggerTest.java
gem-neo4j: Ah yeah, shouldn't be, I just had a lot of issues getting these tests running and was trying everything, removed :)

core/src/test/java/apoc/trigger/TriggerTest.java
gem-neo4j: Good point!

core/src/test/java/apoc/trigger/TriggerTest.java
gem-neo4j: Hmmm I guess it is flaky? but I am not sure if that would have anything to do with my change

common/src/test/java/apoc/util/kernel/MultiThreadedGlobalGraphOperationsTest.java
gem-neo4j: RIP the next commit 😅

core/src/test/java/apoc/trigger/TriggerTest.java
Lojjs: No idea 🤷‍♀️ I guess if it fails on TC we will get a Linear issue, so we can leave it like this for now

Lojjs (approved):
🧹 🥇

PR #890, opened 2026-02-09
Only try to create constraint in CypherInitializer on leader

Lojjs:


gem-neo4j (approved):
Great :) I thought I removed all the sneaky publics in tests 🙈

core/src/main/java/apoc/cypher/CypherInitializer.java
```diff
@@ -118,22 +118,35 @@ public void available() {
...
+                                                .filter(x -> x.getName().equals(constraintName))
+                                                .toList();
+                                        if (maybeConstraint.isEmpty()) {
+                                            tx.schema()
+                                                    .constraintFor(SystemLabels.ApocTriggerMeta)
+                                                    .withName(constraintName)
+                                                    .assertPropertyIsUnique(SystemPropertyKeys.database.name())
+                                                    .create();
+                                            tx.commit();
+                                            constraintCreated = true;
```
gem-neo4j: Should this not be set to true if maybeConstraint exists as well?

Lojjs: Good catch, that would make sense yes

PR #891, opened 2026-02-12
Remove junit4

gem-neo4j:
Removes final imports and references to Junit 5. The removed files are not needed in Core, but are used in Extended, I have message Giuseppe as well as made an Issue on their Github about it, should be easy for them to copy them to Extended.


Lojjs (approved):
Great 🎉 🧹

PR #892, opened 2026-02-13
Use grprc-bom as platform dependency

shaikhu:
Changes necessary for updates to ~arrow-bom~ grpc in neo4j.

Note.
The `PercentilesTest` test failure was due to arrow-bom pulling in newer version of `HdrHistogram` which changed the percentile behaviour e.g. rounding up. Instead use `grpc-bom` version which has the expected behaviour.


JoelBergstrand (approved):
Approved on standup

PR #894, opened 2026-02-17
SURF-646 Replace Guava Sets.intersection() with plain Java in MetaRestricted

neo-harshg:
## Summary
- Replace `com.google.common.collect.Sets.intersection()` with plain Java `HashSet.retainAll()` in `MetaRestricted.java`
- Guava is a `compileOnly` dependency, not bundled in the APOC plugin jar. At runtime it relies on Neo4j providing Guava on its classpath. Newer Neo4j versions may no longer bundle Guava, causing `ClassNotFoundException` for `com.google.common.collect.Sets` when calling `apoc.meta.schema()`
- This eliminates the fragile runtime dependency on Neo4j's internal classpath

https://linear.app/neo4j/issue/SURF-646/test-failure-apocitcoremetacommunityfeaturestestschema


hvub (approved):
Nice.

PR #899, opened 2026-02-19
SURF-600 Fixes [TEST FAILURE] apoc.export.cypher.ExportCypherTest.testIssue2886OptimizationsNoneAndCypherFormatUpdateStructure

neo-harshg:
Fixes  SURF-600

https://linear.app/neo4j/issue/SURF-600/test-failure

**Fixed flaky regex in ExportCypherTest issue 2886 assertions**
The regexMatch helper used \d (single digit) to match node IDs in exported Cypher statements. With newer Neo4j versions, internal node IDs can be multi-digit (>=10), causing the pattern to fail. Changed all \d to \d+ in the issue 2886 test regexes and updated the regexMatch escaping logic to preserve the + quantifier.


JoelBergstrand (approved):
Great!

PR #901, opened 2026-02-23
Fix flakiness of testExportDataGraphML

gem-neo4j:
Looks like the error was due to the ordering, so added an ORDER BY using COLLECT, but as it was so rarely flaky, it is hard to confirm, but that is the woes of APOC :) 

Linear: Fixes SURF-632


JoelBergstrand (approved):
LGTM!

PR #902, opened 2026-02-23
Fix flakiness of testRemoveNode

gem-neo4j:
Add an assertEventually and an await, also remove the negative check, don't think it is needed

Linear: Fixes SURF-631


Lojjs (approved):
Looks good to me

PR #903, opened 2026-02-23
Add error catch and handling to getInputStream

gem-neo4j:
Added some code Giuseppe suggested, and then a test for it. This gives the user a nicer error in the case of a url error instead of just giving a failed procedure error.

Linear: Fixes SURF-267


Lojjs (approved):
I'm not so familiar with the xml tests, so hard to say if the test is perfect but everything here looks OK to me. Nice cleanup of the other tests as well!

PR #909, opened 2026-03-12
Fix wrong error message assertion and improve some assertion outputs

hvub:
It seems the error message in the assertion was simply wrong. There are a few more changes that make assertion outputs on these tests a little more helpful.

### DIFF
--- a/core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
+++ b/core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
@@ -27,7 +27,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 import static org.neo4j.internal.helpers.collection.MapUtil.map;
 import static org.neo4j.test.assertion.Assert.assertEventually;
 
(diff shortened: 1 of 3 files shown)


test-utils/src/main/java/apoc/util/TestUtil.java
```diff
@@ -169,6 +171,47 @@ public static void testFail(GraphDatabaseService db, String call, Class<? extend
...
+            String errorMessage,
+            String nonFailMessage) {
+        try {
+            testCall(db, call, params, r -> fail(nonFailMessage));
+        } catch (RuntimeException e) {
+            assertExceptionMessageContains(e, errorMessage);
+        }
+    }
+
+    public static void assertExceptionMessageContains(RuntimeException e, String errorMessage) {
```
Lojjs, line 192: Nice addition! It is so annoying when a test is flaky and you just get 'expected: true but was: false'

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
```diff
@@ -796,46 +794,37 @@ void testTriggerShow() {
 
     @Test
     void testInstallTriggerInUserDb() {
-        try {
-            testCall(
-                    db,
-                    "CALL apoc.trigger.install('neo4j', 'userDb', 'RETURN 1',{})",
-                    r -> fail("Should fail because of user db execution"));
-        } catch (QueryExecutionException e) {
```
Lojjs, line 804: Some of these tests used to catch the more specific "QueryExecutionException". Are we OK with loosen it to RuntimeException? I assume if we get a custom runtime exception, the error message will be something else then what we are asserting on and the test will fail, so might be fine. 

An option if you want to be more strict is to also send in the expected exception class to the help method, I leave it up to you to decide.

Lojjs (approved):
Looks good, just left an optional change you can choose whether to do

core/src/test/java/apoc/trigger/TriggerNewProceduresTest.java
hvub: Good observation. So, the tests could wrongly succeed — compared to before — if there is an exception of different class thrown but where the message still contains what we assert on. The message assertions are fairly specific so that I find this rather unlikely. If it happens someone can extend the method to also assert on a given exception class.

PR #910, opened 2026-03-17
Do not generate UUIDs in RandomGraphExportTest.

Lojjs:
UUIDs are not yet supported in Cypher so if the db has those they can be exported but not re-parsed in apoc.cypher.runMany again.


gem-neo4j (approved):
great!

PR #911, opened 2026-03-17
Print out line on test failure.

Lojjs:
This assertion seems flaky, but it is not clear what line1 is when it fails and I have not been able to reproduce the failure locally.


JoelBergstrand (approved):
Approved on standup

PR #912, opened 2026-03-17
Fix command injection vulnerabilities in yaml files

Lojjs:
Part of SURF-686

Should be cherry-picked to 5.26

The 4.4 version (which was reported separately) is here: https://github.com/neo4j-contrib/neo4j-apoc-procedures/pull/4533


gem-neo4j (approved):
Nice!

PR #914, opened 2026-03-26
Deprecate apoc.periodic.iterate

gem-neo4j:
CALL {...} IN TRANSACTIONS now covers everything in apoc.periodic.iterate, so we have decided to go ahead with deprecating it in Cypher 25.

Docs: https://github.com/neo4j/docs-apoc/pull/506


core/src/main/java/apoc/periodic/Periodic.java
```diff
@@ -342,14 +342,41 @@ public JobInfo schedule(String name, Runnable task, long delay, long repeat, boo
...
+            }
+            """)
+                    Map<String, Object> config) {
+        return iterate(cypherIterate, cypherAction, config);
+    }
+
+    @Procedure(name = "apoc.periodic.iterate", mode = Mode.WRITE, deprecatedBy = "Cypher's CALL {...} IN TRANSACTIONS.")
+    @Description("Runs the second statement for each item returned by the first statement.\n"
+            + "This procedure returns the number of batches and the total number of processed rows.")
+    @Deprecated
```
Lojjs, line 378: Optional, you can add this to this line `(since = "2026.04", forRemoval = true)`. 

I guess the main benefit is to show that is has been deprecated more than a year and therefore could possibly be removed in Cypher28(or whatever next version is?), but that is also possible to find in the repo history so not sure if it is needed

Lojjs (approved):
Nice! One optional comment based on how I did my last procedure deprecation in Neo4j, but not sure if it is really needed.

PR #915, opened 2026-04-01
SURF-766 Add additionalLabels and additionalProperties config to `apoc.create.virtual.fromNode`

neo-harshg:
Extend the config map of apoc.create.virtual.fromNode to support additionalLabels (LIST<STRING>) and additionalProperties (MAP), allowing users to annotate virtual nodes with extra labels and properties at creation time without requiring write mode.

https://linear.app/neo4j/issue/SURF-766/apoc-add-a-virtual-label-to-a-result-node-to-highlight-eg-potential

### DIFF
--- a/core/src/main/java/apoc/create/Create.java
+++ b/core/src/main/java/apoc/create/Create.java
@@ -411,15 +411,61 @@ public class Create {
     }
 
     @UserFunction("apoc.create.virtual.fromNode")
+    @QueryLanguageScope(scope = {QueryLanguage.CYPHER_5})
     @Description(
             "Returns a virtual `NODE` from the given existing `NODE`. The virtual `NODE` only contains the requested properties.")
-    public Node virtualFromNodeFunction(
+    public Node virtualFromNodeFunctionCypher5(
             @Name(value = "node", description = "The node to generate a virtual node from.") Node node,
             @Name(value = "propertyNames", description = "The properties to copy to the virtual node.")
                     List<String> propertyNames,
             @Name(value = "config", defaultValue = "{}", description = "{ wrapNodeIds = false :: BOOLEAN }")
                     Map<String, Object> config) {
         return new VirtualNode(node, propertyNames, Util.toBoolean(config.get("wrapNodeIds")));
+    }
+
+    @UserFunction("apoc.create.virtual.fromNode")
+    @QueryLanguageScope(scope = {QueryLanguage.CYPHER_25})
+    @Description(
+            "Returns a virtual `NODE` from the given existing `NODE`. The virtual `NODE` only contains the requested properties.")
+    public Node virtualFromNodeFunction(
+            @Name(value = "node", description = "The node to generate a virtual node from.") Node node,
+            @Name(value = "propertyNames", description = "The properties to copy to the virtual node.")
+                    List<String> propertyNames,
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/create/Create.java
```diff
@@ -417,9 +417,33 @@ public Node virtualFromNodeFunction(
...
+        VirtualNode virtualNode = new VirtualNode(node, propertyNames, Util.toBoolean(config.get("wrapNodeIds")));
+
+        Object labelsValue = config.get("additionalLabels");
+        if (labelsValue instanceof List<?> labels) {
+            for (Object l : labels) {
+                if (l instanceof String labelName) {
+                    virtualNode.addLabel(Label.label(labelName));
+                }
+            }
+        }
```
gem-neo4j: I think we should error if the additionalLabels passed is not a LIST<STRING>

core/src/main/java/apoc/create/Create.java
```diff
@@ -417,9 +417,33 @@ public Node virtualFromNodeFunction(
...
+
+        Object propsValue = config.get("additionalProperties");
+        if (propsValue instanceof Map<?, ?> props) {
+            props.forEach((key, value) -> {
+                if (key instanceof String propertyName) {
+                    virtualNode.setProperty(propertyName, value);
```
gem-neo4j, line 463: Same here, fail if wrong type

core/src/test/java/apoc/create/CreateTest.java
```diff
@@ -484,6 +484,101 @@ void testVirtualFromNodeFunctionWithWrapping() {
...
+
+                    assertTrue(node.hasLabel(label("Person")));
+                    assertEquals(1, Iterables.count(node.getLabels()));
+                    assertEquals("Vincent", node.getProperty("name"));
+                    assertNull(node.getProperty("born"));
+                });
+    }
+
+    @Test
+    void testVirtualFromNodeAdditionalPropertyDoesNotOverwriteCopied() {
```
gem-neo4j: The name of the test sounds like it is the opposite of what it is doing?

core/src/main/java/apoc/create/Create.java
```diff
@@ -417,9 +417,33 @@ public Node virtualFromNodeFunction(
             @Name(value = "node", description = "The node to generate a virtual node from.") Node node,
             @Name(value = "propertyNames", description = "The properties to copy to the virtual node.")
                     List<String> propertyNames,
-            @Name(value = "config", defaultValue = "{}", description = "{ wrapNodeIds = false :: BOOLEAN }")
+            @Name(
+                            value = "config",
+                            defaultValue = "{}",
+                            description =
+                                    "{ wrapNodeIds = false :: BOOLEAN, additionalLabels = [] :: LIST<STRING>, additionalProperties = {} :: MAP }")
```
gem-neo4j, line 438: Feature additions can only be made to Cypher 25 versions, so you'll need to duplicate this function and scope it so that the changes are only available there.

gem-neo4j (changes requested):

gem-neo4j (approved):
Great :) Don't forget to write docs here: https://github.com/neo4j/docs-apoc

PR #916, opened 2026-04-02
SURF-789: Add SHA commits to GA

gem-neo4j:
I didn't update the versions even if there are new releases, as I figured "don't touch what ain't broke" :) Just updated them to the SHA as security requested.

Fixes: SURF-789


JoelBergstrand (approved):
Approved on standup

PR #918, opened 2026-04-07
Add semgrep ignore file

gem-neo4j:
This is additive to the one on the platform, it should help us ignore pesky findings that aren't relevant. I have added minimum ones atm, but we can extend


JoelBergstrand (approved):
Approved on standup

PR #919, opened 2026-04-08
Only try to create constraint in CypherInitializer on leader

Lojjs:
Manual backport of https://github.com/neo4j/apoc/pull/890

I skipped to backport the restart test from that PR, because in 5.26 the initializer tests are on community which do not allow start and stop database.


gem-neo4j (approved):
Yep, looks good :) thanks

PR #920, opened 2026-04-09
Allow export lines to be in any order.

Lojjs:
Sometimes the second line is first. This does not make any functional difference,so we can accept both orders in the test to reduce flakiness.

 Part of SURF-701


JoelBergstrand (approved):
LGTM

PR #922, opened 2026-04-16
Deprecate apoc.text.indexOf, apoc.text.join and apoc.text.regexReplace in favour of Cypher.

gem-neo4j:
Cypher has new functions: string.join, string.indexOf and string.regexReplace 🎉 

docs: https://github.com/neo4j/docs-apoc/pull/515


hvub (approved):
No notes.

PR #923, opened 2026-05-04
Fix Cypher Initializer for clusters

gem-neo4j:
Clusters seems to still make it through the is writeable check, as this failing is not actually a problem I just ignore those errors and let it continue. (Also when regenerating license headers this other one popped up, we already have a card for why the CI is passing without checking that)


hvub (approved):
🎉

PR #924, opened 2026-05-04
Fix termination guard in csv loader

gem-neo4j:
The test was flaky depending on when the termination was called and what part was hit, I could reproduce by running the test 20 times in a row before, and after this change that repetition no longer found it :)

### DIFF
--- a/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
+++ b/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
@@ -25,6 +25,7 @@ import apoc.load.CSVResult;
 import apoc.load.Mapping;
 import apoc.load.util.Results;
 import apoc.util.FileUtils;
+import apoc.util.Util;
 import com.opencsv.CSVReader;
 import com.opencsv.CSVReaderBuilder;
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/export/csv/CsvEntityLoader.java
```diff
@@ -133,7 +144,7 @@ public void loadNodes(
                     // we either fail the loading process or skip it depending on the 'ignore duplicate nodes' setting
                     if (idField.isPresent() && idspaceIdMapping.containsKey(nodeCsvId)) {
                         if (clc.getIgnoreDuplicateNodes()) {
-                            return;
+                            continue;
```
JoelBergstrand, line 147: Is this change related to the terminationGuard or seperate?

JoelBergstrand (approved):

core/src/main/java/apoc/export/csv/CsvEntityLoader.java
gem-neo4j: It was in a forEach before and now a while loop, so this is the same behaviour as before (a return in a forEach doesn't exit the whole thing, just that item, same as a continue)

PR #926, opened 2026-05-08
Add IOException throws declaration to CountingInputStream constructor

MishaDemianenko:
The next version of commons IO adds IOException to the BOMInputStream constructor.


gem-neo4j (approved):
Merge on red, it is semgrep complaining and you can ignore that :)

PR #928, opened 2026-05-18
Fix apoc.load.json text by doing a local call instead of online to reduce flakiness

gem-neo4j:
I copied the content of the online call to a local resource. It is always going to be flaky to call random internet websites, so this fixes that.

Linear: resolves SURF-975


JoelBergstrand (approved):
Great!

PR #927, opened 2026-05-15
refactor: add explicit declaration for neo4j-graph-algo dependency

shaikhu:
Needed to ensure PR on `neo-technology` passes.

Add explicit dependency declaration instead of using transitively. 

## Proposed Changes (Mandatory)

A brief list of proposed changes in order to fix the issue:

  - add explicit declaration for `neo4j-graph-algo` dependency.

PR #929, opened 2026-05-18
build: add declaration for neo4j-graph-algo dependency

shaikhu:
Follow up to previous [PR](https://github.com/neo4j/apoc/pull/927). Dependency needs to be added to `core` module as well. 😅


gem-neo4j (approved):
LGTM :)

PR #930, opened 2026-05-21
Fix apoc.load.csv by checking for EOF instead of null bytes

gem-neo4j:
The csv file we were calling in the IT tests was a url that got a bot check added to it, the bot check returns different info than other failing cases, and it got stuck in a while loop looking for a null value instead of EoF.

### DIFF
--- a/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
+++ b/core/src/main/java/apoc/export/csv/CsvEntityLoader.java
@@ -336,7 +336,7 @@ public class CsvEntityLoader {
     private static String readFirstLine(CountingReader reader) throws IOException {
         String line = "";
         int i;
-        while ((i = reader.read()) != 0) {
+        while ((i = reader.read()) != -1) {
             char c = (char) i;
             if (c == '\n') break;
             line += c;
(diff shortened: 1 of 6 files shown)


core/src/main/java/apoc/export/csv/CsvEntityLoader.java
```diff
@@ -336,7 +336,7 @@ private Map<String, Mapping> getMapping(List<CsvHeaderField> fields) {
     private static String readFirstLine(CountingReader reader) throws IOException {
         String line = "";
         int i;
-        while ((i = reader.read()) != 0) {
+        while ((i = reader.read()) != -1) {
```
JoelBergstrand, line 339: Is it only AWS that is using `0` for the bot check? I assume this line applies to any provider, e.g. does this hold for other providers as well?

gem-neo4j: This is to do with checking for the null value or EoF, EoF is the correct way to check if a file is finished, not really sure why this was done, I assume it was a mistake, and this is how Neo4j checks for CSV file ending, hence why LOAD CSV ends with 0 rows on this url instead of running forever.

JoelBergstrand (approved):

core/src/main/java/apoc/export/csv/CsvEntityLoader.java
JoelBergstrand: I see! Great!

PR #932, opened 2026-05-26
Clean up apoc.algo.*

gem-neo4j:
My goal is to use Claude to help improve the APOC repo. The idea being to go by group of callable, and do the following:

* Pull in apoc.common implementation into core, so it is easier for us to drop common eventually
* Add edge case testing
* Improve code readability and reduce things like dead code etc
* Update the [docs](https://github.com/neo4j/docs-apoc/pull/526) for each as I go

### DIFF
--- a/core/src/main/java/apoc/algo/CorePathFindingUtils.java
+++ b/core/src/main/java/apoc/algo/CorePathFindingUtils.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) "Neo4j"
+ * Neo4j Sweden AB [http://neo4j.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


core/src/main/java/apoc/algo/CorePathFindingUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import org.neo4j.graphdb.Direction;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.PathExpander;
+import org.neo4j.graphdb.PathExpanderBuilder;
+import org.neo4j.graphdb.RelationshipType;
+import org.neo4j.values.storable.PointValue;
+
+public class CorePathFindingUtils {
+    public static class GeoEstimateEvaluatorPointCustom implements EstimateEvaluator<Double> {
+        // -- from org.neo4j.graphalgo.impl.util.GeoEstimateEvaluator
```
JoelBergstrand, line 33: This is from Neo4j and not common if I understand correctly?

JoelBergstrand (approved):
I think it looks nice - things like merging the strings are nice from a readability perspective.

core/src/main/java/apoc/algo/CorePathFindingUtils.java
gem-neo4j: Well the file is a direct copy from Common, to start (well restart I guess) the work of stopping using common. But yeah, I think originally it would've been

PR #934, opened 2026-06-01
Add missing dep

gem-neo4j:
This should fix the startup test issues, but as those are tested using docker, they won't actually go green until after this is merged 😓


JoelBergstrand (approved):
Approved on standup

Issue #933, opened 2026-05-29
Zero/inaccurate relationships count from apoc.meta.schema


Shady-Hegazy:
For the below bug report, we exemplify using the Geography DB from the CypherBench collection of datasets.[Schema file.](https://huggingface.co/datasets/megagonlabs/cypherbench/blob/main/graphs/schemas/geography_schema.json)

## Expected Behavior

apoc.meta.schema() and apoc.meta.schema({sample: -1}) should return accurate relationship count values per node label, where:

- Outgoing counts are non-zero when relationships exist.
- Source-side (out) and target-side (in) counts for the same directed relationship are equal.
- Per-label-pair counts reflect only relationships between that specific label pair, not global totals.

## Actual Behavior

Both commands produce inaccurate count values. Four problems observed:

1. Outgoing counts consistently reported as 0 for high-cardinality source labels (in Geography domain e.g., Mountain, River, Lake) even when the target-side incoming count is non-zero. It seems to occur exclusively for outgoing relationships.  
2. Consequently: source/target counts are contradictory for the same relationship (e.g., flowsThrough: River out → 0, Country in → 82,155).
3. Per-label-pair counts equal global totals — the same global count is reported for multiple distinct target labels simultaneously (e.g., both Lake.flowsInto and Ocean.flowsInto report 75,139).
4. Sampled mode (without 'sample:-1')severely undercounts (e.g., flowsInto: sampled 1,032 vs full-scan 75,139).

## How to Reproduce the Problem

1. Load and start a dataset that has directed relationships.
2. Run CALL apoc.meta.schema({sample: -1})
3. Compare reported counts for 'OUT' relationships against actual counts in the DB. Plus, check the duplication (reuse of global total counts) for relationships that are defined for different node pairs.
4. Run CALL apoc.meta.schema() and compare sampled vs full-scan counts 

### Simple Dataset (where it's possibile)

Geography dataset in https://github.com/megagonlabs/cypherbench. Neo4j docker images are available.

### Steps (Mandatory)

1. Load and start a dataset that has directed relationships.
2. Run CALL apoc.meta.schema({sample: -1})
3. Compare reported counts for 'OUT' relationships against actual counts in the DB. Plus, check the duplication (reuse of global total counts) for relationships that are defined for different node pairs.
4. Run CALL apoc.meta.schema() and compare sampled vs full-scan counts 

## Specifications

Version: 2026.04.0
APOC version: 2026.04.0-core
[...]

PR #935, opened 2026-06-01
SURF 1040: Fix Zero/inaccurate relationships count from apoc.meta.schema

gem-neo4j:
Fix the bug that made outgoing relationship counts be set to zero. See https://github.com/neo4j/apoc/issues/933

### DIFF
--- a/core/src/main/java/apoc/meta/MetaRestricted.java
+++ b/core/src/main/java/apoc/meta/MetaRestricted.java
@@ -759,7 +759,7 @@ public class MetaRestricted {
                                         "direction",
                                         "out",
                                         "count",
-                                        metaItem.rightCount,
+                                        metaItem.leftCount,
                                         "labels",
                                         metaItem.other,
                                         "properties",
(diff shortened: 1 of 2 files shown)


core/src/main/java/apoc/meta/MetaRestricted.java
```diff
@@ -759,7 +759,7 @@ private Map<String, Object> collectNodesMetaData(
                                         "direction",
                                         "out",
                                         "count",
-                                        metaItem.rightCount,
+                                        metaItem.leftCount,
```
JoelBergstrand, line 762: Do I understand correctly that the `rightCount` represents the incoming relationships and the `leftCount` the outgoing? There by the bug as this is `out`.

JoelBergstrand (approved):

core/src/main/java/apoc/meta/MetaRestricted.java
gem-neo4j: Yeah that was my understanding, where they are defined they have this:

```
leftCount += out;
rightCount += in;
```

PR #936, opened 2026-06-04
Fix apoc.search.multiSearchReduced UnsupportedOperationException

gem-neo4j:
When writing new examples I found that the search procedures fail very easily when anything tries to mutate the list/map as they were unmutable. There were no tests for some of these at all, so probably not so strange they didn't work 😅


hvub (approved):
🎉

PR #939, opened 2026-06-05
Add IOException throws declaration to CountingInputStream constructor…

ericsporre:
Cherry picks (https://github.com/neo4j/apoc/pull/926) since we 5.26 has also bumped the dependency of apache commons configuration2.


gem-neo4j (approved):
LGTM

PR #941, opened 2026-06-23
Updated assertion in MetaCommunityFeaturesTest.schema

JoelBergstrand:
Assertion should have been updated together with  [SURF 1040: Fix Zero/inaccurate relationships count from apoc.meta.schema
](https://github.com/neo4j/apoc/commit/8275283f7253e4e28cafcacff659039a41f7f66c) but for some reason wasn't.


hvub (approved):
`MetaCommunityFeaturesTest.schema` is lists as successful in the TC log on this PR, so I assume the test ran and the fix is correct.

PR #945, opened 2026-07-22
Close cursor-backed resources to prevent leaks at commit 

JoelBergstrand:
Wrap relationship/node iterables and query Results in try-with-resources
(and keep onClose on procedure-returned streams) so kernel cursors are
released on early exit or exception. Adds regression tests.

### DIFF
--- a/common/src/main/java/apoc/export/cypher/formatter/CypherFormatterUtils.java
+++ b/common/src/main/java/apoc/export/cypher/formatter/CypherFormatterUtils.java
@@ -32,6 +32,7 @@ import org.neo4j.graphdb.Direction;
 import org.neo4j.graphdb.Label;
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.Relationship;
+import org.neo4j.graphdb.ResourceIterable;
 import org.neo4j.values.storable.DurationValue;
 import org.neo4j.values.storable.Value;
 import org.neo4j.values.storable.Values;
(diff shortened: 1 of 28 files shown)


core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -287,11 +286,12 @@ void testMetaType() {
...
-                    .evaluator(toDepth(1))
-                    .traverse(node)
-                    .iterator()
-                    .next();
+            // This iterator is not closed properly and is unused
+            // Path path = tx.traversalDescription()
+            //         .evaluator(toDepth(1))
+            //         .traverse(node)
+            //         .iterator()
+            //         .next();
```
gem-neo4j: Maybe delete this then?

core/src/test/java/apoc/meta/MetaTest.java
```diff
@@ -327,11 +327,12 @@ void testMetaIsType() {
...
```
gem-neo4j: delete?

gem-neo4j (approved):
Great, just a small comment about deleting dead code :)

core/src/test/java/apoc/meta/MetaTest.java
JoelBergstrand: Should I leave the old TODO or remove that as well?

`            // TODO PATH FAILS              testTypeName(path, "PATH");`

gem-neo4j: May as well remove it, if it is something broken we either should have a card or someone in the future will make a bug report

PR #948, opened 2026-08-06
SURF-1259: Exclude vector arrays

gem-neo4j:
Exclude vector arrays from random graph test


hvub (approved):
Surprisingly many dependency bumps necessary to exclude vector arrays ;-)

PR #950, opened 2026-08-17
build: use assignment syntax for Gradle DSL properties

shaikhu:
First of batch of PRs to resolve gradle deprecation warnings (make upgrade to Gradle 9 possible)

Gradle 8.13 deprecates the Groovy DSL space-assignment (`propName value`) and method-call property syntax (`propName(value)`). 

Convert the six places we used them.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ allprojects {
       mavenCentral()
     }
     maven {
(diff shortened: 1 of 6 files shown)


gem-neo4j (approved):
Ooh thanks :) I've been meaning to do this sometime 😅

PR #952, opened 2026-08-17
SURF-1348:Trigger duplicate metadata nodes bug

gem-neo4j:
If the constraint has failed and somehow there are 2 metadata trigger nodes, introduce a recovery step of deleting them and updating the remaining.


hvub (approved):
no notes.

PR #956, opened 2026-08-18
build: replace archivesBaseName with base.archivesName

shaikhu:
Third batch of updates to remove gradle deprecation warnings.


gem-neo4j (approved):
thanks

PR #962, opened 2026-09-07
SURF-1480: Clean up more open cursors; subgraphAll

gem-neo4j:
I couldn't reproduce the actual issue, but this may help, I have also contacted Kernel to see if they can make changes on their side.


JoelBergstrand (approved):
Looks good to me

PR #965, opened 2026-09-08
SURF-1403: APOC security issue; Archive entry fully decompressed into memory before size limit is applied

gem-neo4j:
Moved the LimitedSizeInputStream wrap to bound bytes as they're read from the archive entry itself, before IOUtils.toByteArray runs (see card for recommended fix for more details)


fredrik-neo4j (approved):

JoelBergstrand (approved):
So was the logic there already just in the wrong ordered?

gem-neo4j:
@JoelBergstrand wrongly ordered for one use case, it worked for other use cases (hence the other tests in the file), but yeah, pretty much :P

PR #966, opened 2026-09-08
Cypher injection fixes

gem-neo4j:
Fixes a bunch of potential cypher injection cases:

Linear: Fixes SURF-1414, SURF-1413, SURF-1411, SURF-1410, SURF-1409, SURF-1404


fredrik-neo4j (approved):
LGTM!

core/src/main/java/apoc/merge/Merge.java
```diff
@@ -276,7 +280,7 @@ public Stream<RelationshipResultWithStats> relationshipWithStatsEager(
     private String buildIdentPropsString(Map<String, Object> identProps) {
         if (identProps == null) return "";
         return identProps.keySet().stream()
-                .map(Util::quote)
+                .map(key -> Util.sanitize(key, true))
```
JoelBergstrand, line 283: Is there a practical use case for `Util.quote` or should it be removed in favor for `Util.sanitize`

JoelBergstrand (approved):

core/src/main/java/apoc/merge/Merge.java
JoelBergstrand: Seems to have 50 uses so might be that easy of a replacement

gem-neo4j: yeah, didn't really want to open that can of worms 😅

PR #967, opened 2026-09-09
SURF-1412: Fix Decompression-bomb limit disabled when Content-Length is absent, chunked, or exceeds 2GB

gem-neo4j:
* Fixes silent truncating from using getLength (int) instead of getContentLengthLong (long).
* Adds a new config item for max size of 2gb (see attached docs)
* No longer trusts the servers declared content length and instead checks this and makes sure it isn't too large

See docs: https://github.com/neo4j/docs-apoc/pull/549


common/src/test/java/apoc/export/util/LimitedSizeInputStreamTest.java
```diff
@@ -0,0 +1,123 @@
...
+                    limited.read(buf);
+                }
+            });
+            assertTrue(e.getMessage().contains("compression bomb"));
+        }
+
+        // it must have been stopped after consuming only a handful of "source" bytes - proving the
+        // budget tracks real consumption instead of a declared length, which would otherwise have
+        // let all 10_000 MB of bogus output through unchecked.
+        assertTrue(consumedSourceBytes.get() < 10);
```
hvub, line 121: It is a bit unclear to me where the configured limits are set here, so that stop the loop before 10 iterations. The DEFAULT_MAX_DECOMPRESSION_SIZE of 2Gb would be too high for this test to succeed, because it would need 1.908 1MB reads to exceed 2Gb. It seems to get limited by the DEFAULT_MAX_DECOMPRESSION_RATIO = 200, right? The ratioBound is recomputed after every read and it 1:1MB, which is clearly high than min(1 * 200B, 2Gb). That should already happen after the first consumed source byte, right?

Anyway, I am just poking it see if I understand the test logic correctly.

hvub (approved):
No real notes.

fredrik-neo4j (approved):

common/src/test/java/apoc/export/util/LimitedSizeInputStreamTest.java
gem-neo4j: Yeah, this test is more about making sure we test real consumption, and not just what we are told the size will be (declared length aka Content-Length), and like you thought, it should be hitting the ratio logic

PR #968, opened 2026-09-10
Surf 1405: More Cypher injection clean up

gem-neo4j:
I followed Joels comment on my last PR and decided to update everything to use the safer way of sanitizing, but APOC extended uses the old way, so I left it for them with a comment.

Linear: fixes: SURF-1405, SURF-1436


common/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
```diff
@@ -80,40 +81,48 @@ public String statementForNodeIndex(
...
     @Override
     public String statementForNodeFullTextIndex(String name, Iterable<Label> labels, Iterable<String> keys) {
         String label = StreamSupport.stream(labels.spliterator(), false)
                 .map(Label::name)
-                .map(Util::quote)
+                .map(Util::quoteIdentifierSafely)
                 .collect(Collectors.joining("|"));
         String key = StreamSupport.stream(keys.spliterator(), false)
-                .map(s -> String.format(PROPERTY_QUOTING_FORMAT, "n", s))
+                .map(s -> String.format(PROPERTY_QUOTING_FORMAT, "n", s.replace("`", "``")))
```
JoelBergstrand: Is there a particular reason this uses a hand crafted escape? Same with the one further down.

common/src/main/java/apoc/util/Util.java
```diff
@@ -692,10 +692,19 @@ public static void sleep(int millis) {
...
     }
 
+    // This is in common, so I suspect this is used by extended, leaving as is, please use quoteIdentifierSafely
+    // instead.
     public static String quote(String var) {
         return SourceVersion.isIdentifier(var) && !var.contains("$") ? var : '`' + var + '`';
     }
 
+    public static String quoteIdentifierSafely(String value) {
+        if (SourceVersion.isIdentifier(value) && !value.contains("$")) {
```
JoelBergstrand, line 703: Unrelated to your change but this `!var.contains("$")` I guess is some kind of parameter quick escape? E.g. you don't want to quote the param `$parameter`. If that's the case - won't it behave weirdly with input that has `$` in the variable name? `var$name`?

JoelBergstrand (approved):

common/src/main/java/apoc/util/Util.java
```diff
@@ -692,10 +692,19 @@ public static void sleep(int millis) {
         }
     }
 
+    // This is in common, so I suspect this is used by extended, leaving as is, please use quoteIdentifierSafely
+    // instead.
```
fredrik-neo4j, line 696: is there a way to warn us when invoking this method? perhaps an annotation saying that its deprecated with a pointer to the quoteIdentifierSafely method. it might not be possible since the quote method seem to be used elsewhere, and is not in fact considered deprecated

common/src/main/java/apoc/util/Util.java
gem-neo4j: So, the SourceVersion.isIdentifier doesn't include dollar signs, so it checks if the string contains that and then quotes it. So var$name would always be quoted which is correct behaviour

common/src/main/java/apoc/util/Util.java
gem-neo4j: I can add a deprecated, the reason it stays is because the common module is shared by the APOC extended repo, so if I remove it they may get failures there all of a sudden :)

common/src/main/java/apoc/export/cypher/formatter/AbstractCypherFormatter.java
gem-neo4j: Yeah true, should probably use the sanitize function :) will update

common/src/main/java/apoc/util/Util.java
fredrik-neo4j: i think that makes sense, its not a blocker for me so you can decide. just wanted to point out that there might a chance that its overlooked ( :

gem-neo4j: I have already added it :P would you like to approve the PR if you're happy?

fredrik-neo4j (approved):

common/src/main/java/apoc/util/Util.java
fredrik-neo4j: oh my bad, it was my intention to write this comment & approve PR at the same time!

PR #972, opened 2026-09-22
SURF-1406: Make user-supplied regexes in apoc.text terminable

JoelBergstrand:
Matcher never checks for interruption and Strings held no `TerminationGuard`, so a backtracking match could not be stopped by `dbms.killQuery` or by transaction timeout.

Evaluate user regexes against a CharSequence wrapper that checks the guard every 10,000 `charAt` calls. Matcher reads through `charAt` on every backtracking step, so this makes any match abortable. subSequence returns the delegate's subsequence unwrapped, so results stay plain strings.

Applied to `regexGroups`, `regexGroupsByName`, `regreplace` (backing both `apoc.text.replace` overloads) and split. split now narrows its limit with `Math.toIntExact`, so a limit above `Integer.MAX_VALUE` errors instead of silently behaving as limit = 0.

The termination tests use (x+x+)+y, which backtracks cubically. The `(a+)+$` shape usually quoted here is pruned by the JDK and completes in under a millisecond, so a test built on it would prove nothing.


hvub (approved):
Sweet.

PR #973, opened 2026-09-23
SURF-1581: Report failures from streaming exports

JoelBergstrand:
Streaming exports run the dump on a pool thread and end the stream with a tombstone posted from a finally block, so a dump that threw looked exactly like one that succeeded: the Future holding the exception was discarded, and the caller saw an empty or truncated result with no error.

Keep the Future and rethrow its failure when the consumer reaches the tombstone. This exposed a second bug it had been hiding: ProgressReporter ends the dump with ProgressInfo.EMPTY, which is null, but the consumers compared against their own non-null EMPTY constant and threw a NullPointerException on every streaming export. Treat null as the terminal marker.

Move the queue, executor and spliterator wiring into ExportUtils.streamProgressInfo so apoc.export.cypher shares one copy.


gem-neo4j (approved):
Ah the ever going battle with the streams and tombstones 😅 this change looks good

